PR #5, opened 2019-10-15
better english, fix error in steps

zeroheure:


product_pack/readme/USAGE.rst
```diff
@@ -1,10 +1,10 @@
 To use this module, you need to:
 
-#. Go to *Sales > Products > Products*, create a product and set "Is Pack?".
-#. Then set the products has to include in the product pack of the product.
+#. Go to *Sales > Products > Products*, create a product and set *Is Pack?*.
```
pedrobaeza: Don't change italics here.

product_pack/readme/USAGE.rst
```diff
@@ -1,10 +1,10 @@
...
 #. Set *Pack Type* and *Pack component price*.
+#. Then choose the products to include in the pack.
 
-`Product pack` is a base module for `sale_product_pack` and others modules that
+Product pack` is a base module for `sale_product_pack` and other modules that
```
pedrobaeza: Remove both quotes or any, but not half. You can enclose it by double quotes or put its technical name.

pedrobaeza (changes requested):

product_pack/readme/USAGE.rst
zeroheure: Done, but IMHO italic is easier to read, because it is used previously (I come from typography :-)

product_pack/readme/USAGE.rst
zeroheure: Done.

pedrobaeza (approved):
Thanks. Fast-tracking. I said about not putting in italics because that's a field label, not a menu entry (defined through a path), and then I prefer to differentiate both through the way they are highlighted.

PR #12, opened 2020-03-18
[12.0][IMP] sale_product_pack add pack_modifiable in the sale order line.

hparfr:
It's helpfull to know directly if the sale order line is editable.

If the sale order is edited from an api, onchange are not always triggered
so having this field helps to raise an error.

And it can be used to display in the view that the line is not editable.

### DIFF
--- a/sale_product_pack/models/sale_order_line.py
+++ b/sale_product_pack/models/sale_order_line.py
@@ -30,6 +30,8 @@ class SaleOrderLine(models.Model):
         'pack_parent_line_id',
         'Lines in pack'
     )
+    pack_modifiable = fields.Boolean(
+        help='The parent pack is modifiable')
 
     @api.multi
     def expand_pack_line(self, write=False):
@@ -88,7 +90,18 @@ class SaleOrderLine(models.Model):
         """ Do not let to edit a sale order line if this one belongs to pack
         """
         if self._origin.pack_parent_line_id and \
-           not self._origin.pack_parent_line_id.product_id.pack_modifiable:
+           not self._origin.pack_modifiable:
             raise UserError(_(
                 'You can not change this line because is part of a pack'
                 ' included in this order'))
+
+    @api.multi
+    def action_open_parent_pack_product_view(self):
+        domain = [('id', 'in', self.mapped(
+            'pack_parent_line_id').mapped('product_id').ids)]
+        return {'name': _('Parent Product'),
+                'type': 'ir.actions.act_window',
+                'res_model': 'product.product',
+                'view_type': 'form',
+                'view_mode': 'tree,form',
+                'domain': domain}
--- a/sale_product_pack/models/product_pack_line.py
+++ b/sale_product_pack/models/product_pack_line.py
@@ -22,6 +22,7 @@ class ProductPack(models.Model):
             'pack_parent_line_id': line.id,
             'pack_depth': line.pack_depth + 1,
             'company_id': order.company_id.id,
+            'pack_modifiable': line.product_id.pack_modifiable,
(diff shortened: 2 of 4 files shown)


PierrickBrun (approved):
Code review

ernestotejeda:
I think that being modifiable or not is a characteristic of the product class, not of the order line class.
Image we try to modify a SOline and we get an exception and then we realize that the product has that option marked incorrectly, when we uncheck it to fix it we will have to delete and create the SOLine for this change to take effect, or remove the product and select it again.
What do you think @pedrobaeza ?
Wouldn't it be better to have a related field instead?

hparfr:
>  Image we try to modify a SOline and we get an exception and then we realize that the product has that option marked incorrectly, when we uncheck it to fix it we will have to delete and create the SOLine for this change to take effect, or remove the product and select it again.

The product can be in other sale order as well. Or even multiple times in the same sale order and It will be not updated automatically.

pedrobaeza:
Why the new field is not shown on the UI?

hparfr:
> Why the new field is not shown on the UI?
Because there is already a lot of fields on the sale order lines views.
For the moment I prefer letting the integrator to decide if he wants to display it and how (with color ? a logo?...)

pedrobaeza:
Then I don't see any advantage on having the field ...

hparfr:
> Then I don't see any advantage on having the field ...

- It's easier to call `sale_order_line.is_modifiable` than `sale_order_line.pack_parent_line_id.product_id.pack_modifiable`
- Better perfs

pedrobaeza:
That performance is not so critical in this case, taking into account that is not continuously checked and for sure will be in cache. Any other reason?

hparfr:
Be more pythonic is not a reason good enough ?

I don't see drawbacks of having it as a field.

(BTW, I use it on e-commerce shop to know if the guy is allowed to change quantities on the line and to manage properly the unlink).

pedrobaeza:
I prefer to provide whole set in the same PR (Odoo UI part).

hparfr:
Added the field in the view, just for you @pedrobaeza

pedrobaeza (changes requested):
Sorry to still be picky, but with this change, a migration script is required for existing sale order lines with pack that moves the value from product template to line. Can you please include it?

sale_product_pack/models/sale_order_line.py
```diff
@@ -88,7 +90,18 @@ def check_pack_line_modify(self):
...
         """
         if self._origin.pack_parent_line_id and \
-           not self._origin.pack_parent_line_id.product_id.pack_modifiable:
+           not self._origin.pack_modifiable:
             raise UserError(_(
                 'You can not change this line because is part of a pack'
                 ' included in this order'))
+
+    @api.multi
+    def action_openParentPackProductView(self):
```
pedrobaeza: Please use always snake case for method variables:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#342variable-names

hparfr:
> Sorry to still be picky, but with this change, a migration script is required for existing sale order lines with pack that moves the value from product template to line. Can you please include it?

Yes you are ;-).  I don't think it's worth the effort. Having some existing draft sale order without the flag is ok and there is no value of changing confirmed sale orders.

> Please use always snake case for method variables

Fixed; thanks

pedrobaeza:
OK, please squash commits and I'll merge.

hparfr:
Thanks @pedrobaeza

pedrobaeza (approved):

PR #13, opened 2020-03-26
[FIX] fixes mysterious CacheMiss errors

gbaumgartner-callino:
I've got some CacheMiss Errors using product_pack module while additionally using sale.order.template.

Steps were:
1) creating a new Order with a Template
2) deleting all predefined Lines
3) Searching for a new product
4) CacheMiss Error happens at search

So while debugging i found that changing the time when the super call would happen somehow fixed the issue.

I don't see any problems with changing the super call position so if anyone else might run into the same problem this PR might fix it for them too.


pedrobaeza:
This PR is changing the behavior of the compute, as the pack price is not respected, so this is incorrect. You must look for another fix.

PR #15, opened 2020-04-13
[12.0] [FIX] product_pack : use sudo in the get_pack_lines method

nicomacr:
This is because in the e-commerce when an user portal or public try to access to the catalog of products has an access error to the model "product pack line".
 Them problem is when the method price_compute is call then try to compute the products expand the product pack and try to access to that model.
This way evoid to set an access  read to "product pack line" for the portal and public users.

### DIFF
--- a/product_pack/models/product_product.py
+++ b/product_pack/models/product_product.py
@@ -47,7 +47,7 @@ class ProductProduct(models.Model):
                 lambda p: p.pack_ok and p.pack_type == 'detailed'
                 and p.pack_component_price == 'detailed')
 
-        no_packs = (self | self.get_pack_lines().mapped('product_id')) - packs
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I think you should do the `sudo` where that method is called, for isolating possible security leaks. Any way, please include detailed steps to reproduce the problem to value the best solution to this.

nicomacr:
@pedrobaeza i've just  correct the position of sudo to do in the the method called.
To reproduce the issue follow this steps:

1. Create a  product pack with any of type of packs.
2. Go to the front-end (website) page, go to the shop page, an select any product to buy.
3. You will se the error for access model.

zaoral (approved):

PR #18, opened 2020-04-29
[FIX] stock_product_pack: only use stockable products to calculate the stock.

azucenatrey:


pedrobaeza:
Thanks for the PR. Merging as the previous one is approved:

/ocabot merge patch

PR #21, opened 2020-05-12
[12.0][FIX][sale_product_pack] Allow to update product_uom_qty after sale is confirmed

### LINKED ISSUE #20: [sale_product_pack] can't update quantity when sale is not draft
With module sale_product_pack

In a sale order with state = draft, when you change the ordered quantity of the parent product, the child product is updated.

But, when sale is confirmed, updating the ordered quantity of the parent product do not update the child product anymore.

@ernestotejeda 
What is the reason for this behavior ?
https://github.com/OCA/product-pack/blame/12.0/sale_product_pack/models/sale_order_line.py#L42

hparfr:
Fixes #20

### DIFF
--- a/sale_product_pack/models/sale_order_line.py
+++ b/sale_product_pack/models/sale_order_line.py
@@ -39,7 +39,6 @@ class SaleOrderLine(models.Model):
         do_not_expand = self._context.get('update_prices') or \
             self._context.get('update_pricelist', False)
         if (
-                self.state == 'draft' and
                 self.product_id.pack_ok and
                 self.pack_type == 'detailed'):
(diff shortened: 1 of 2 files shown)


PierrickBrun (approved):
Yes, I don't think this is the place to choose if you can edit the so line or not when the so is done

pedrobaeza:
How this works with a linked picking? Everything is properly updated?

hparfr:
If the quantities are decreased (in sale) it does nothing in the picking.
If the quantities are increased (in sale) the picking get the new quantities for all the lines.

pedrobaeza (approved):
OK, seems correct then

PR #23, opened 2020-06-07
[13.0][MIG] product_pack

jjscarafia:

PR #24, opened 2020-06-07
[WIP][13.0][MIG] sale_product_pack

jjscarafia:
This PR needs this one to be merged before: https://github.com/OCA/product-pack/pull/23

I've an error while deleting pack lines from SO and I couldn't fix it. Seams related to changes on orm v13


sale_product_pack/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
...
+        _origin.order_line only when lines are unlinked and this is exactly
+        what we need
+        """
+        if self._origin.order_line.filtered(
+            lambda x: x.pack_parent_line_id
+            and not x.pack_parent_line_id.product_id.pack_modifiable
+        ):
+            raise UserError(
+                _(
+                    "You can not delete this line because is part of a pack in"
```
ernestotejeda:
```suggestion
                    "You cannot delete this line because is part of a pack in"
```

sale_product_pack/models/sale_order_line.py
```diff
@@ -0,0 +1,109 @@
...
+        "name",
+        "tax_id",
+    )
+    def check_pack_line_modify(self):
+        """ Do not let to edit a sale order line if this one belongs to pack
+        """
+        if self._origin.pack_parent_line_id and not self._origin.pack_modifiable:
+            raise UserError(
+                _(
+                    "You can not change this line because is part of a pack"
```
ernestotejeda, line 93:
```suggestion
                    "You cannot change this line because is part of a pack"
```

pedrobaeza:
Please rebase

sale_product_pack/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
...
+        pack_copied_lines.unlink()
+        return sale_copy
+
+    @api.onchange("order_line")
+    def check_pack_line_unlink(self):
+        """At least on embeded tree editable view odoo returns a recordset on
+        _origin.order_line only when lines are unlinked and this is exactly
+        what we need
+        """
+        if self._origin.order_line.filtered(
```
ernestotejeda: Now with odoo 13, self._origin.order_line is not the deleted line, but the remaining lines as new lines (NewId). So, It can be a posible solution. ping @pedrobaeza 

```suggestion
        origin_line_ids = self._origin.order_line.ids
        line_ids = self.order_line.ids
        removed_line_ids = list(set(origin_line_ids) - set(line_ids))
        removed_line = self.env['sale.order.line'].browse(removed_line_ids)
        if removed_line.filtered(
```

ernestotejeda (changes requested):

jjscarafia:
hi @ernestotejeda @pedrobaeza 
1. I've applied the rebase
2. I've fix the can not for cannot
3. I've applied the Ernesto recomendation to fix the unlink on the onchange but I still can't fix the pack deletion properly. Check [this video](https://drive.google.com/file/d/1fhOD7dcRMtq4xqptEGuKqIuWmHcnmOvj/view)

sale_product_pack/models/sale_order_line.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # Fields for common packs
+    pack_depth = fields.Integer(
+        "Depth", help="Depth of the product if it is part of a pack."
+    )
+    pack_parent_line_id = fields.Many2one(
+        "sale.order.line",
+        "Pack",
+        help="The pack that contains this product.",
+        ondelete="cascade",
```
ernestotejeda, line 21:
```suggestion
```

sale_product_pack/models/sale_order_line.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def unlink(self):
+        """Remove previously the pack children lines for avoiding issues in
+        the cache.
+        """
+        children = self.mapped("pack_child_line_ids")
+        if children:
+            children.unlink()
+        return super().unlink()
+
```
ernestotejeda, line 77:
```suggestion

```

sale_product_pack/models/sale_order.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def copy(self, default=None):
+        sale_copy = super().copy(default)
+        # we unlink pack lines that should not be copied
+        pack_copied_lines = sale_copy.order_line.filtered(
+            lambda l: l.pack_parent_line_id.order_id == self
+        )
+        pack_copied_lines.unlink()
+        return sale_copy
+
```
ernestotejeda, line 18:
```suggestion
    def write(self, vals):
        if 'order_line' in vals:
            to_delete_ids = [e[1] for e in vals['order_line'] if e[0] == 2]
            sub_packs = self.env['sale.order.line'].search(
                [('pack_parent_line_id', 'in', to_delete_ids)])
        res = super().write(vals)
        if sub_packs:
            sub_packs_command = [(2, c.id, 0) for c in sub_packs]
            self.write({'order_line': sub_packs_command})
        return res

```

ernestotejeda (commented):
What I have found is the following.
Now in odoo 13, when a line is removed from the sales order form, Odoo take the remaining lines as lines to add. So, if at the same time those remaining lines are removed by code, like we do, then Odoo exploits here https://github.com/OCA/OCB/blob/6e56f1ccacc075eb94b9e2b179e0762b0221e2f6/odoo/fields.py#L2996-L2996

I have overridden the sale.order write method to try to solve the problem there. check the next suggested changes related with this and tell me what do you think?
ping @pedrobaeza

sale_product_pack/models/sale_order.py
pedrobaeza: Yes, I would suggest something similar, but try to include in original `vals` the `[(2, ...)]` commands for doing all in one step.

jjscarafia:
Thanks you both for the tips
@ernestotejeda I believe you are more familiar with this change and the proper way to fix it. If you want to make a new PR with those fixes it's ok for me. If you prefer I will try to fix it, but I believe you will achieve it really faster :)

ernestotejeda:
It's ok, I'll do it :+1:

PR #27, opened 2020-07-08
[13.0][MIG] sale_product_pack: Migration to 13.0

ernestotejeda:
supersedes https://github.com/OCA/product-pack/pull/24

cc @Tecnativa TT24634


ernestotejeda:
@jjscarafia could you review?.

pedrobaeza:
Please squash administrative commits together and several translations that comes together from the same person.

pedrobaeza:
@ernestotejeda I don't see your commit with extra changes. Have you included in @jjscarafia's one? They should be split.

ernestotejeda:
Yes, I included my changes and the same commit. But, ok, I will divide into 2 commits

ernestotejeda:
@pedrobaeza I have squashed administrative commits and I split the v13 migration commit into 2 commits.

jjscarafia (approved):
Tested and working ok. +1 for code review also
thanks for the great work @ernestotejeda 
In the future it would be great a cleaner / better approach for the lines unlink, but for now this is enough for me

sale_product_pack/models/sale_order.py
```diff
@@ -37,3 +37,19 @@ def check_pack_line_unlink(self):
...
+            to_delete_ids = [e[1] for e in vals["order_line"] if e[0] == 2]
+            subpacks_to_delete_ids = (
+                self.env["sale.order.line"]
+                .search(
+                    [("id", "child_of", to_delete_ids), ("id", "not in", to_delete_ids)]
+                )
+                .ids
+            )
+            if subpacks_to_delete_ids:
+                for cmd in vals["order_line"]:
```
pedrobaeza, line 52: What happens if one of the `subpacks_to_delete_ids` is not in the command line? That one won't get removed. You need to assure that all needed `(2, x)` commands are there.

pedrobaeza (changes requested):

ernestotejeda:
@pedrobaeza , done. Check that.

pedrobaeza (approved):

jjscarafia:
thanks you both!

PR #28, opened 2020-09-08
[FIX] sale_product_pack: don't relay on demo data

chienandalu:
In CIs with default localizations we could get a pricelist with an
unexpected currency

cc @Tecnativa TT25413


pedrobaeza:
The text doesn't seems correct.

chienandalu:
What do you mean?

pedrobaeza:
Previously there was no pricelist modifications :smile:

pedrobaeza (approved):

victoralmau (approved):

PR #37, opened 2021-02-25
[13.0] [IMP] product_pack: Add rights access in online quotations.

nicomacr:
If you use a pack in online quotations, it won't let you see it due to permissions with the lines when you send the budget to a public / Portal user.

### DIFF
--- a/product_pack/models/product_product.py
+++ b/product_pack/models/product_product.py
@@ -59,7 +59,7 @@ class ProductProduct(models.Model):
         )
         for product in packs.with_context(prefetch_fields=False):
             pack_price = 0.0
-            for pack_line in product.pack_line_ids:
+            for pack_line in product.sudo().pack_line_ids:
                 pack_price += pack_line.get_price()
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
It's not correct to add any permission globally. Please specify a group, and add proper record rules according such group.

nicomacr:
@pedrobaeza i've add the proper groups for each record rules.

pedrobaeza:
I still see that you are giving too much permission. This is a security breach, showing all your pack lines to anyone via XML-RPC. Maybe you need to add a glue module for online quotations, or intercept the call at code level, but giving such permission is not suitable.

nicomacr:
@pedrobaeza i understand your concern about the security breach.  I've already made a different approach to solve this problem.
Check this out in the commit.

pedrobaeza (approved):
I don't see any risk here, but I want to know the opinion of @ernestotejeda and @chienandalu

chienandalu (approved):
I think it's ok. Thanks!

PR #40, opened 2021-04-12
[13.0] [IMP] stock_product_pack: Not add the product pack to the move…

nicomacr:
…s if the product is storable.

### DIFF
--- a/stock_product_pack/models/product_template.py
+++ b/stock_product_pack/models/product_template.py
@@ -1,0 +1,16 @@
+# Copyright 2019 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    dont_create_move = fields.Boolean(
+        string="Don't create move",
+        help="With this option, the pack won't create an stock.move and will"
+        " be set as delivered upon sale confirmation. This is useful to use "
+        "get pack stock availability (type = 'product') but"
+        " without actually having stock and moves of it.",
+    )
--- a/stock_product_pack/models/stock_rule.py
+++ b/stock_product_pack/models/stock_rule.py
@@ -1,0 +1,23 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class ProcurementGroup(models.Model):
+    _inherit = "procurement.group"
+
+    @api.model
+    def run(self, procurements):
+        """ If 'run' is called on a pack product storable.
+         we remove the procurement with this product pack.
+        """
+        for procurement in procurements:
+            if (
+                procurement.product_id
+                and procurement.product_id.pack_ok
+                and procurement.product_id.dont_create_move
+                and procurement.product_id.pack_type == "detailed"
+            ):
+                procurements.remove(procurement)
+
+        return super().run(procurements)
--- a/stock_product_pack/tests/test_stock_product_pack.py
+++ b/stock_product_pack/tests/test_stock_product_pack.py
@@ -15,6 +15,7 @@ class TestSaleProductPack(SavepointCase):
         super().setUpClass()
         category_all_id = cls.env.ref("product.product_category_all").id
         cls.product_obj = cls.env["product.product"]
+        cls.stock_rule_obj = cls.env["stock.rule"]
         component_1 = cls.product_obj.create(
             {"name": "Component 1", "type": "product", "categ_id": category_all_id}
         )
@@ -30,6 +31,37 @@ class TestSaleProductPack(SavepointCase):
                 "name": "Pack",
                 "type": "product",
                 "pack_ok": True,
+                "pack_type": "detailed",
+                "pack_component_price": "detailed",
+                "categ_id": category_all_id,
+                "pack_line_ids": [
+                    (0, 0, {"product_id": component_1.id, "quantity": 1},),
+                    (0, 0, {"product_id": component_2.id, "quantity": 1},),
+                    (0, 0, {"product_id": component_3.id, "quantity": 1},),
+                    (0, 0, {"product_id": component_4.id, "quantity": 1},),
+                ],
+            }
+        )
+        warehouse = cls.env["stock.warehouse"].search(
+            [("company_id", "=", cls.env.user.id)], limit=1
+        )
+        cls.stock_rule = cls.stock_rule_obj.create(
+            {
+                "name": "Stock to Costumers",
+                "action": "pull",
+                "picking_type_id": cls.env.ref("stock.picking_type_internal").id,
+                "route_id": cls.env.ref("stock.route_warehouse0_mto").id,
+                "procure_method": "make_to_stock",
+                "warehouse_id": warehouse.id,
+                "location_id": cls.env.ref("stock.stock_location_stock").id,
+            }
+        )
+        cls.pack_dc_with_dm = cls.product_obj.create(
+            {
+                "name": "Pack With storeable and not move product",
+                "type": "product",
+                "pack_ok": True,
+                "dont_create_move": True,
                 "pack_type": "detailed",
                 "pack_component_price": "detailed",
                 "categ_id": category_all_id,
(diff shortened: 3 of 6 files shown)


jjscarafia:
@pedrobaeza any thoughts about it?
As explained on the code, the idea is to have stock of products packs.

More info:

- Odoo only compute stock for product types in "consu" / "product"
- But then, on the views, only if type "product" is shown.
- Also, on ecommerce, the stock is only considered if type = 'product'
- with this changes we can use pack type = 'product' to get benefit of using stock but without actually moving the packs as products in stock

pedrobaeza:
Not sure about this. Maybe a better approach would be to not hide such buttons if the product is a pack, touching the conditions on the smart-button visibility. I fear this will have side effects in places that you haven't though about.

@chienandalu what is your opinion on this?

chienandalu (commented):
I'd say a better approach would be something like this module that solves the same issue with kits: https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_available_mrp

jjscarafia:
Thanks you both for the quick response.

If I'm not wrong the "stock_available_mrp" improves for products that are produced and would be of type "product", so odoo already will shown the stock info for this products. The module is improving how this quantity is computed

This is similar already done in product pack here https://github.com/OCA/product-pack/blob/12.0/stock_product_pack/models/product_product.py#L11

The problems are:
1. odoo is showing stock fields only if product type = 'product'
2. odoo, since v13, is not computing stock for products type = 'service' (https://github.com/odoo/odoo/blob/13.0/addons/stock/models/product.py#L101)
3. odoo is using stock on ecommerce only if product type = 'product': https://github.com/odoo/odoo/blob/13.0/addons/website_sale_stock/static/src/xml/website_sale_stock_product_availability.xml#L6
https://github.com/odoo/odoo/blob/9b91f9d1bb41d05f445d745a9b8e5fa63edaf2a9/addons/website_sale_stock/models/sale_order.py#L18
https://github.com/odoo/odoo/blob/9b91f9d1bb41d05f445d745a9b8e5fa63edaf2a9/addons/website_sale_stock/static/src/js/variant_mixin.js#L49

So, if you want to use packs as a "phantom bom" but detailing products on the sale order (and not needing mrp), the most common use case for packs I believe, you can't actually have the stock of the product. If you:
* set pack as service: stock is not computed, shown and not being considered on ecommerce
* set pack as consu: stock is computed, not shown and not being considered on ecommerce. Also, pack will travel to stock.picking and this is not always wanted
* set pack as producto: stock is computed, shown on views and is being considered on ecommerce. Pack will travel to stock.picking and this is not always wanted.

With this new little **option** we can make it work

New thoughts?

chienandalu:
> If I'm not wrong the "stock_available_mrp" improves for products that are produced and would be of type "product", so odoo already will shown the stock info for this products. The module is improving how this quantity is computed

Well, it works with consu as well:
image


> set pack as consu: stock is computed, not shown and not being considered on ecommerce. Also, pack will travel to stock.picking and this is not always wanted

That would be a blocking point indeed. And it's one hard to solve indeed (views to be modified in core and possible modules extending the core ones).

> set pack as producto: stock is computed, shown on views and is being considered on ecommerce. Pack will travel to stock.picking and this is not always wanted.

Maybe a good approach would be to do it as with phantom products:
https://github.com/odoo/odoo/blob/14.0/addons/mrp/models/stock_move.py#L223

The kit is exploded and after that the original move is removed.

stock_product_pack/models/stock_rule.py
```diff
@@ -0,0 +1,28 @@
...
+                and procurement.product_id.pack_ok
+                and procurement.product_id.dont_create_move
+                and procurement.product_id.pack_type == "detailed"
+            ):
+                procurements.remove(procurement)
+                sale_order_id = procurement.values.get("sale_line_id")
+                if sale_order_id:
+                    self.env["sale.order.line"].browse(
+                        sale_order_id
+                    ).qty_delivered = procurement.product_qty
```
pedrobaeza: This is triggering the invoiceability of the line before it should, so it's not a valid option.

pedrobaeza:
@chienandalu it can't be done the same as kits, as the whole goal of this module is to still show the pack on the picking, not to "hide" it through its components.

stock_product_pack/models/stock_rule.py
jjscarafia: you're right Pedro, it would affect if invoice policy is "delivered".
Maybe we can try to make the same approach as boms here, regarding the delivery of the components this quantity would be updated. We try to avoid so because it was more complex but I didn't think about the invoiceability.

What do you think? Any other suggestion?

pedrobaeza: I still think that changing the visibility of the smart-buttons should be enough, isn't it?

jjscarafia: I don't think so:
a) if you use consumable: stock will be visible, the pack would still travel to the picking and you will move a product that is not a product. Also, on ecommerce it wont consider stock (so actually you would sell packs of products without stock). It's heavier to fix that on ecomerce:
https://github.com/odoo/odoo/blob/13.0/addons/website_sale_stock/static/src/xml/website_sale_stock_product_availability.xml#L6
https://github.com/odoo/odoo/blob/9b91f9d1bb41d05f445d745a9b8e5fa63edaf2a9/addons/website_sale_stock/models/sale_order.py#L18
https://github.com/odoo/odoo/blob/9b91f9d1bb41d05f445d745a9b8e5fa63edaf2a9/addons/website_sale_stock/static/src/js/variant_mixin.js#L49

b) if you use it a service: stock could be visible (we also need to patch the method as proposed [here](https://github.com/OCA/product-pack/pull/41), same issues on ecommerce and the product wont be set as delivered automatically. You would need to put manually the delivered quantity

pedrobaeza: Then what is your suggestion?

jjscarafia: My suggestion is go for this pr with two options:
1. keep as is but adding a warning or known issue that if the user decides too use the new option "dont_create_move" with products with invoice policy "to deliver", then the "invoiceability" won't be accurate.
2. improve this PR and try to set the delivery quantity from the delivered pack lines, similar to what is done from product phantom boms

For me we can start with "1" and keep improvement on the road-map, but if you prefer "2" we can go for that.

What do you think?

pedrobaeza: For me 1 is not acceptable. Let's go for 2.

chienandalu:
I added this module to cover the delivery qtys computing https://github.com/OCA/product-pack/pull/47

I think you could remove that part (and the dependencies change) from this PR and just keep the part removing the pack from the picking move by configuration

nicomacr:
@chienandalu excellent news, yes i've keep only the part of the skip the product pack for going to the stock moves.

chienandalu (approved):
A test case would be very welcome

stock_product_pack/tests/test_stock_product_pack.py
```diff
@@ -89,3 +106,39 @@ def test_compute_quantities_dict(self):
         self.product_obj.invalidate_cache()
         self.assertEqual(self.pack_dc.virtual_available, 5)
         self.assertEqual(self.pack_dc.qty_available, 5)
+
+    def test_pack_with_dont_move_the_parent(self):
+        """ Run a procurement for prod pack products when there are only 5 in stock then
+        check that MTO is applied on the moves when the rule is set to 'mts_else_mto'
```
chienandalu, line 127: Oh, forget it. I was relaying on the `sale` dependency... Maybe triggering the procurement with a stock orderpoint?


<strike>
Why you don't simply create a sale order with the detailed pack and confirm it? It's as simple as using the form class:

```python
self.pack_dc.dont_create_move = True
sale_form = Form(self.env["sale.order"])
sale_form.partner_id = self.env["res.partner"].create({"name": "Mr. Odoo"})
with sale_form.order_line.new() as line:
    line.product_id = self.pack_dc
sale = sale_form.save()
sale.action_confirm()
# And now do the assertions
self.assertFalse(self.pack_dc in sale.picking_ids.move_lines.product_id)
```

You can create another test with the oposite case as well
</strike>

nicomacr: Yes i realised that dependency of sale is not available here for this module. I've try some ideas, but they didn't work.
Any suggest to do it using stock orderpoint?

chienandalu: The simplest way I come up with would be to create this rule and then run the scheduler (after configuring the pack product as storable and setting up an orderpoint):

image

nicomacr:
@chienandalu i'ts ready to review.

chienandalu (approved):

chienandalu:
Thanks!

stock_product_pack/models/product_template.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    dont_create_move = fields.Boolean(
+        string="Don't create move",
+        help="With this option the pack won't create and stock.move and will"
```
pedrobaeza:
```suggestion
        help="With this option, the pack won't create an stock.move and will"
```

stock_product_pack/models/stock_rule.py
```diff
@@ -0,0 +1,23 @@
...
+        for procurement in procurements:
+            if (
+                procurement.product_id
+                and procurement.product_id.pack_ok
+                and procurement.product_id.dont_create_move
+                and procurement.product_id.pack_type == "detailed"
+            ):
+                procurements.remove(procurement)
+
+        return super(ProcurementGroup, self).run(procurements)
```
pedrobaeza:
```suggestion
        return super().run(procurements)
```

nicomacr:
@pedrobaeza ready this changes!

pedrobaeza (approved):

PR #42, opened 2021-04-29
[13.0] [FIX] product_pack: Use sudo when calculate the price and the user doesn't have access to write in the model product_product.

nicomacr:
Step to reproduce the issue:

Configuration to replicate the scenario with the bug:
 Select  a user without any administration rights, either sales, invoice or accounting. I use the "demo" user in runbot to replicate.

1) In the list view, if we search in the bar for the name of some product pack, an error appears.

The error is like that:

```
Sorry, you are not allowed to modify documents of type 'Product' (product.product). This operation is allowed for the groups:
	- Administration/Settings
	- Inventory/Administrator
	- Invoicing/Billing Administrator
	- Sales/Administrator - (Operation: write, User: 6)
```

This also happens when add a product in the sale line.

### DIFF
--- a/product_pack/models/product_product.py
+++ b/product_pack/models/product_product.py
@@ -98,7 +98,7 @@ class ProductProduct(models.Model):
         to_uom = None
         if "uom" in self._context:
             to_uom = self.env["uom.uom"].browse([self._context["uom"]])
-        for product in packs:
+        for product in packs.sudo():
             list_price = product.price_compute("list_price").get(product.id)
             if to_uom:
                 list_price = product.uom_id._compute_price(list_price, to_uom)


pedrobaeza:
Which are the steps to reproduce the problem?

nicomacr:
@pedrobaeza I update the comment in the PR adding the part to replicate the problem.

pedrobaeza:
I have just tried in runbot in http://3480114-13-0-b53c3f.runbot2.odoo-community.org, and I'm not able to reproduce the problem. You miss in your steps the creation of the pack, but I even create a product called "Pack", and still no error. What are the missing steps to reproduce the problem?

nicomacr:
@pedrobaeza I tested in that runbot and i replicated de issue.
I show it to you in this [video](https://drive.google.com/file/d/1Of4uTZQiu5VvO5Y6dnfQCEIBfJuK2444/view) with demostration.

pedrobaeza:
Which of the packs is provoking the problem? The steps to reproduce should include such information (Create a product, mark the check "Is pack?", put this and there...), and it should be part of the commit message.

nicomacr:
@pedrobaeza with any configurations of packs this problems is happens.
I updated the commit message adding more information for the problem.

nicomacr:
@pedrobaeza what do you think about this change?
I've complete the commit message with more details.

pedrobaeza (approved):
I still don't find the problem, but don't worry more about it, I have checked if there can be side effects of using sudo there, and I don't find any, so let's go with this.

bruno-zanotti (approved):
LGTM

PR #46, opened 2021-06-18
[ADD] website_sale_product_pack: New module

chienandalu:
Compatibility module between sale_product_pack and website_sale

cc @Tecnativa TT30385

### DIFF
--- a/website_sale_product_pack/__manifest__.py
+++ b/website_sale_product_pack/__manifest__.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Website Sale Product Pack",
+    "category": "E-Commerce",
+    "summary": "Compatibility module of product pack with e-commerce",
+    "version": "13.0.1.0.0",
+    "license": "AGPL-3",
+    "depends": ["website_sale", "sale_product_pack"],
+    "data": ["views/assets.xml", "views/templates.xml"],
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/product-pack",
+    "installable": True,
+}
--- a/website_sale_product_pack/models/sale_order.py
+++ b/website_sale_product_pack/models/sale_order.py
@@ -1,0 +1,49 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _cart_update(
+        self, product_id=None, line_id=None, add_qty=0, set_qty=0, **kwargs
+    ):
+        """We need to keep the discount defined on the components when checking out"""
+        line = self.env["sale.order.line"].browse(line_id)
+        if line and line.pack_parent_line_id:
+            return super(
+                SaleOrder, self.with_context(pack_discount=line.discount)
+            )._cart_update(product_id, line_id, add_qty, set_qty, **kwargs)
+        return super()._cart_update(product_id, line_id, add_qty, set_qty, **kwargs)
+
+    @api.depends("order_line.product_uom_qty", "order_line.product_id")
+    def _compute_cart_info(self):
+        """We only want to count the main pack line, not the component lines"""
+        super()._compute_cart_info()
+        for order in self:
+            order.cart_quantity = int(
+                sum(
+                    order.website_order_line.filtered(
+                        lambda x: not x.pack_parent_line_id
+                    ).mapped("product_uom_qty")
+                )
+            )
+
+    def _website_product_id_change(self, order_id, product_id, qty=0):
+        """In the final checkout step, we could miss the component discount as the
+        product prices are recomputed"""
+        res = super()._website_product_id_change(order_id, product_id, qty=qty)
+        if self.env.context.get("pack_discount"):
+            res["discount"] = self.env.context.get("pack_discount")
+        return res
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def unlink(self):
+        """The website calls this method specifically. We want to get rid of
+        the children lines so the user doesn't have to"""
+        join_pack_children = self + self.mapped("pack_child_line_ids")
+        return super(SaleOrderLine, join_pack_children.exists()).unlink()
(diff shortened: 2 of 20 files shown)


website_sale_product_pack/controllers/main.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import http
+
+from odoo.addons.website_sale.controllers.main import WebsiteSale
+
+
+class WebsiteSaleProductPack(WebsiteSale):
+    @http.route()
+    def product(self, product, category="", search="", **kwargs):
+        res = super().product(product, category=category, search=search, **kwargs)
+        http.request.env.context = dict(http.request.env.context, whole_pack_price=True)
```
pedrobaeza: Please add docstring about this. I think there's a way of not doing Python hack (maybe through `qcontext`?)

website_sale_product_pack/models/sale_order.py
```diff
@@ -0,0 +1,37 @@
...
+        """We need to keep the discount defined on the components when checking out"""
+        line = self.env["sale.order.line"].browse(line_id)
+        if line and line.pack_parent_line_id:
+            return super(
+                SaleOrder, self.with_context(pack_discount=line.discount)
+            )._cart_update(product_id, line_id, add_qty, set_qty, **kwargs)
+        return super()._cart_update(product_id, line_id, add_qty, set_qty, **kwargs)
+
+    def _website_product_id_change(self, order_id, product_id, qty=0):
+        """In the final checkout stop, we could miss the component discount as the
```
pedrobaeza:
```suggestion
        """In the final checkout step, we could miss the component discount as the
```

website_sale_product_pack/models/sale_order.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def unlink(self):
+        """The website calls this method specifically. We want to get rid of
+        the children lines so the user doesn't have to"""
+        for line in self.filtered("pack_child_line_ids"):
+            self |= line.pack_child_line_ids
```
pedrobaeza: Don't add over `self`, but have a new variable (`obj` for example), and call super with it. You can as well have it directly:
```suggestion
        obj = self + self.mapped("pack_child_line_ids")
```

website_sale_product_pack/models/sale_order.py
chienandalu: fixed. thanks :)

website_sale_product_pack/models/sale_order.py
chienandalu: fixed :+1:

website_sale_product_pack/controllers/main.py
chienandalu: It didn't work as expected anyway. I'm removing it

website_sale_product_pack/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Website Sale Product Pack",
+    "category": "E-Commerce",
+    "summary": "Compatibility module of product pack with e-commerce",
+    "version": "13.0.1.0.0",
+    "license": "AGPL-3",
+    "depends": ["website_sale", "sale_product_pack"],
```
pedrobaeza, line 9: Shouldn't this depend on `website_sale_stock`?

website_sale_product_pack/views/assets.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template id="assets_frontend" inherit_id="website_sale.assets_frontend">
```
pedrobaeza: Isn't there specific assets for tests? Or at least load it on demo.

chienandalu: Yes. Done

website_sale_product_pack/__manifest__.py
chienandalu: A pack of services would be valid as well

chienandalu:
@ernestotejeda could you review?

pedrobaeza (approved):

pedrobaeza:
Ey, Travis goes to red now

chienandalu:
:green_apple:

nicomacr:
@chienandalu great work with this module. Really thanks
We got a similar module [here](https://github.com/ingadhoc/website/tree/13.0/website_product_pack) 

I've check this PR and compare with our module and find that would be interesting to have this features in the module to handle the publish mechanism:

- [here](https://github.com/ingadhoc/website/blob/13.0/website_product_pack/models/product_product.py#L13) we check for the products to add in a pack that all of them were published y if it not we alert. **Maybe with a domain in the field could be better.**

- [here](https://github.com/ingadhoc/website/blob/13.0/website_product_pack/models/product_template.py#L12) we set for the pack lines in a pack to be publish or not from the parent pack. And the inverse case if one of the pack component becomes to unpublish we set the same for the parent pack. **I think this part is better made in another method calling from write method.**
 
I thinks this could be necessary because the mixing of a product pack published with somes of the components not, could be confuse and some bug related to this will be happened.


Any thoughts about that?


Thanks again.

chienandalu:
Thanks @nicomacr I propose you to add your review in order to merge this one, and then I'll gladly review your PR with those improvements :slightly_smiling_face:

nicomacr (approved):
LGTM

pedrobaeza:
@chienandalu rebase to see the problem.

alvaro-domatix:
We've detected a problem trying to use the packs on the website.

If you configure a product as a pack with pack_type = "detailed" and pack_component_price = "ignored" it apparently works while you do the cart until you try to pay. Once you enter the address information and click on "Next", the order sums the amount of all the components and the pack too.

Example:

Pack 1 (30€):
- Comp 1 (20€)
- Comp 2 (20€)
Once you click on next after the address information the order's amount is 30€ + 20€ + 20€ = 70€

website_sale_product_pack/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
+This module introduces compatibility of product packs with e-commerce:
+
+- In the cart summary, the components aren't editable and are shown hanging
+  from the main pack reference.
+- When we remove a pack from the cart, their components are removed as well.
+- The cart popup summary only shows the main pack line and discards the sublines in
+  the units count.
+- The cart summary shows the component lines hanging from the main one as well.
+- It's ensured the the prices are shown correctly for the whole pack and that they're
```
ernestotejeda, line 9:
```suggestion
- It's ensured the prices are shown correctly for the whole pack and that they're
```

ernestotejeda (commented):
I have functionally tested the module and I have found the same error that @alvaro-domatix  mentions here https://github.com/OCA/product-pack/pull/46#issuecomment-926829089
Also, when a pack is removed from the cart, its components are not removed.

PR #49, opened 2021-09-02
[13.0] [FIX] sale_stock_product_pack: avoid error when the cancel quantity for all components.

nicomacr:
In case you cancel the quantity for all components of the pack an error occurred, after this change we set the quantity delivery to zero.


nicomacr:
@chienandalu could you check this PR?

jjscarafia (approved):
+1

chienandalu (approved):

nicomacr:
@pedrobaezai think this PR wasn't merge because is pending an approval (green one) of a maintainer.

pedrobaeza:
There's a problem with the ocabot. @sbidoul is also broken like OCA instance?

nicomacr:
@pedrobaeza it's fixed the problems with ocabot?

PR #51, opened 2021-09-28
[13.0][FIX] stock_product_pack: Remove unnecessary and wrong action from tests

joao-p-marques:
The `stock.scheduler.compute` wizard action is run in a different Thread that the main one.
In a test environment, that is problematic, as the test thread ends without actually getting the result from the 2nd that is launched.
This can lead to unclosed cursors and an inconsistent state in the environment.
Besides, for the same reason, we can simply delete the step from the test, as it was not doing anything.

@Tecnativa
TT32016

ping @pedrobaeza


pedrobaeza (approved):

pedrobaeza:
Fast-tracking for fixing the problem in integration tests:

/ocabot merge patch

PR #52, opened 2021-10-13
[13.0][FIX] Don't allow to modify price_unit of a pack_line if its parent pack_component_price is 'totalized' or 'ignored'.

alvaro-domatix:
Basically, this is a fix to solve the problem that happens when you try to register the payment on the website after completing the address information.

### DIFF
--- a/sale_product_pack/i18n/ca.po
+++ b/sale_product_pack/i18n/ca.po
@@ -6,15 +6,15 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 12.0\n"
 "Report-Msgid-Bugs-To: \n"
-"PO-Revision-Date: 2019-12-27 14:13+0000\n"
-"Last-Translator: Carles Antoli <carlesantoli@hotmail.com>\n"
+"PO-Revision-Date: 2022-02-28 13:17+0000\n"
+"Last-Translator: Noel estudillo <noelestudilloviera@gmail.com>\n"
 "Language-Team: none\n"
 "Language: ca\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: \n"
 "Plural-Forms: nplurals=2; plural=n != 1;\n"
-"X-Generator: Weblate 3.10\n"
+"X-Generator: Weblate 4.3.2\n"
 
 #. module: sale_product_pack
 #: model:ir.model.fields,field_description:sale_product_pack.field_sale_order_line__pack_depth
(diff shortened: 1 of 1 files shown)


chienandalu (commented):
Did you see our PR https://github.com/OCA/product-pack/pull/46 solving and extending compatibility with the website?

ernestotejeda (commented):
Hello @alvaro-domatix.
Lines with components of a detail-ignored or detailed-totalized pack are created with unit price set to 0 (https://github.com/OCA/product-pack/blob/13.0/sale_product_pack/models/product_pack_line.py#L38) and there is a method to prevent the price and other fields to be changed (https://github.com/OCA/product-pack/blob/13.0/sale_product_pack/models/sale_order_line.py#L77), so, i don't think this change is necessary. Could you please tell me, a use case in which we need this change? thanks.

alvaro-domatix:
I saw it and I commented the error I found [there](https://github.com/OCA/product-pack/pull/46#issuecomment-926829089). Although I have only detected the error using the website, it's because a method of sale_product_pack that is called one extra time when you do the order on the website that things go wrong.

I made this patch and it works but I'm not telling is the best way to solve it. If you want to see the error you have to install the website_sale_product_pack #46 and buy a pack with pack_type = "detailed" and pack_component_price = "ignored" from the website. Apparently goes well until you complete the invoicing address and you click on "next" to choose payment method.

alvaro-domatix:
Can we merge this? @pedrobaeza

pedrobaeza:
I need the confirmation from my colleagues that this change is ok, but I see that they don't consider as needed.

sale_product_pack/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "category": "Sales",
     "summary": "This module allows you to sale product packs",
     "website": "https://github.com/OCA/product-pack",
-    "author": "NaN·tic, ADHOC SA, Tecnativa, Odoo Community Association (OCA)",
+    "author": "NaN·tic, ADHOC SA, Tecnativa, Domatix, Odoo Community Association (OCA)",
```
dreispt: A fix does not grant authorship.
Please add yourself to the contributors list instead.

rousseldenis:
@pedrobaeza  @alvaro-domatix Maybe through an option ?

sale_product_pack/__manifest__.py
pedrobaeza: This is pending

pedrobaeza (changes requested):
It can be put through an option, but it's not considered as something needed by my colleagues

rousseldenis:
> It can be put through an option, but it's not considered as something needed by my colleagues

The question is not what is needed by your colleagues but more what is common to all OCA users.

PR #54, opened 2021-10-20
[13.0][FIX] sale_product_pack: Avoid error in tests with other installed addons

victoralmau:
Add rate to the budget to avoid inconsistent prices in some cases..

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT32539

### DIFF
--- a/sale_product_pack/tests/test_sale_product_pack.py
+++ b/sale_product_pack/tests/test_sale_product_pack.py
@@ -8,8 +8,29 @@ class TestSaleProductPack(SavepointCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        pricelist = cls.env["product.pricelist"].create(
+            {
+                "name": "Test",
+                "company_id": cls.env.company.id,
+                "item_ids": [
+                    (
+                        0,
+                        0,
+                        {
+                            "applied_on": "3_global",
+                            "compute_price": "formula",
+                            "base": "list_price",
+                        },
+                    )
+                ],
+            }
+        )
         cls.sale_order = cls.env["sale.order"].create(
-            {"partner_id": cls.env.ref("base.res_partner_12").id}
+            {
+                "company_id": cls.env.company.id,
+                "partner_id": cls.env.ref("base.res_partner_12").id,
+                "pricelist_id": pricelist.id,
+            }
         )
 
     def _get_component_prices_sum(self, product_pack):


sale_product_pack/tests/test_sale_product_pack.py
```diff
@@ -8,8 +8,15 @@ class TestSaleProductPack(SavepointCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        main_company = cls.env.ref("base.main_company")
```
joao-p-marques: Why does this always have to be associated with `base.main_company`? In that case, maybe ensure `self.env.company` is that main company...

pedrobaeza: I agree

sale_product_pack/tests/test_sale_product_pack.py
```diff
@@ -8,8 +8,15 @@ class TestSaleProductPack(SavepointCase):
...
+        pricelist = cls.env["product.pricelist"].create(
+            {"name": "Test", "currency_id": main_company.currency_id.id}
```
pedrobaeza: This is missing the minimal pricelist item for using the sales price.

victoralmau:
Changes done.
It is necessary to create a pricelist and define it in the order to avoid incorrect prices (the default pricelist that is defined in some cases has a different currency and defined "discounts" that change the price).

sale_product_pack/tests/test_sale_product_pack.py
```diff
@@ -8,8 +8,12 @@ class TestSaleProductPack(SavepointCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        pricelist = cls.env["product.pricelist"].create({"name": "Test"})
```
pedrobaeza: Please force the company:
```suggestion
        pricelist = cls.env["product.pricelist"].create({"name": "Test", "company_id": cls.env.company.id})
```

sale_product_pack/tests/test_sale_product_pack.py
```diff
@@ -8,8 +8,12 @@ class TestSaleProductPack(SavepointCase):
...
         cls.sale_order = cls.env["sale.order"].create(
-            {"partner_id": cls.env.ref("base.res_partner_12").id}
+            {
+                "partner_id": cls.env.ref("base.res_partner_12").id,
```
pedrobaeza, line 31: The same here:
```suggestion
                "company_id": cls.env.company.id,
                "partner_id": cls.env.ref("base.res_partner_12").id,
```

pedrobaeza:
I'm still missing one pricelist item inside the new pricelist for the default rule of list_price.

sale_product_pack/tests/test_sale_product_pack.py
victoralmau: A line is already added in the pricelist linked to the product that is used later.

sale_product_pack/tests/test_sale_product_pack.py
```diff
@@ -8,8 +8,32 @@ class TestSaleProductPack(SavepointCase):
...
     def setUpClass(cls):
         super().setUpClass()
+        cls.product_cp = cls.env.ref(
+            "product_pack.product_pack_cpu_detailed_components"
+        )
+        pricelist = cls.env["product.pricelist"].create(
+            {
+                "name": "Test",
+                "company_id": cls.env.company.id,
+                "item_ids": [
```
pedrobaeza, line 15: Well, I was talking about the generic line for all products with list price...

victoralmau: Applying a rule of a fixed price to all products that use that pricelist will cause other parts of this test to fail.

Although the initial problem was in the wrong price in some cases, it was due to the fact that a "general" pricelist was being applied that changed the price in a specific case and also in another currency. In this case, creating only the rate without rules already solves the problem and lets the price of the product sheet apply.

pedrobaeza: But I'm not saying to apply a fixed price rule. All time I'm talking about "list price". Please open a runbot and see the generic pricelist to see what I'm talking about. It's very simple. You just have to open Odoo, not to see code.

victoralmau: Ok sorry, I already understood what you meant.

pedrobaeza (approved):

joao-p-marques (approved):

PR #55, opened 2021-11-11
[13.0] [IMP] website_sale_product_pack: Published and unpublished for products using packs

nicomacr:
For keep the consistent and prevent bugs within the e-commerce, we force that all childs of a parent pack
stay publish when the parent is published.
Also if any of the childs of the parent pack became unpublish, we unpublish the parent.

### DIFF
--- a/website_sale_product_pack/models/product_template.py
+++ b/website_sale_product_pack/models/product_template.py
@@ -1,0 +1,40 @@
+# Copyright 2019 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    @api.constrains("is_published")
+    def check_website_published(self):
+        """ For keep the consistent and prevent bugs within the e-commerce,
+         we force that all childs of a parent pack
+         stay publish when the parent is published.
+         Also if any of the childs of the parent pack became unpublish,
+         we unpublish the parent."""
+        for rec in self.filtered(lambda x: x.pack_ok and x.is_published):
+            unpublished = rec.pack_line_ids.mapped("product_id").filtered(
+                lambda p: not p.is_published
+            )
+            if unpublished:
+                raise ValidationError(
+                    _("You can't unpublished products (%s) to a published " "pack (%s)")
+                    % (", ".join(unpublished.mapped("name")), rec.name)
+                )
+        for rec in self.filtered(
+            lambda x: not x.is_published and x.used_in_pack_line_ids
+        ):
+            published = rec.used_in_pack_line_ids.mapped("parent_product_id").filtered(
+                "is_published"
+            )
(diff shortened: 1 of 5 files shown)


nicomacr:
@chienandalu here some improvements to this module, what do you think?

website_sale_product_pack/models/product_template.py
```diff
@@ -0,0 +1,34 @@
...
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    def write(self, vals):
+        """ For keep the consistent and prevent bugs within the e-commerce,
+         we force that all childs of a parent pack
+         stay publish when the parent is published.
+         Also if any of the childs of the parent pack became unpublish,
+         we unpublish the parent."""
+        if len(vals) == 1 and "website_published" in vals and vals["website_published"]:
```
chienandalu: Why `len(vals) == 1`

nicomacr: because if came more values i don't want to change for all the records(parent childs) the other values. This way i ensure that only change the `website published` field not other fields.

Are you agree with that or have another idea?

nicomacr: @chienandalu

nicomacr:
@chienandalu ping

nicomacr:
@rousseldenis what do you think about this changes?

nicomacr:
@rousseldenis what do you think about this changes?
@chienandalu

nicomacr:
@ernestotejeda could you check this PR?

website_sale_product_pack/models/product_template.py
chienandalu: For the children case, I'd just throw a warning. For the parents case, it'd be ok. You can update the parents anyway so you don't need to restrict the length of the vals. Otherwise you could be missing some cases.

nicomacr: I've just change the approach of this. 
When publish a pack, we publish all of the child's also.
When unpublish one of the child's, we unpublish the parent product

chienandalu: > When publish a pack, we publish all of the child's also.

I think that approach is problematic and could create a chain of conflicts when those pack components are in several packs with different publishing states. We should just throw a warning to avoid publishing such pack and let the user review the state.

> When unpublish one of the child's, we unpublish the parent product

Not as conflicting, but the user could not be aware of such behavior. Maybe the warning could be the way to go again.

nicomacr: Yes I agree with you, i've made the changes in the commit.

pedrobaeza:
Please rebase and avoid the merge commits in the PR, and let's finish it.

pedrobaeza:
The commit name is weird.

nicomacr:
> The commit name is weird.

I've changed the name to represent the change in this PR

pedrobaeza (approved):
Seems OK. The copyright in the files is not the correct though.

chienandalu (approved):
LGTM :+1: Change attribution to yourself anyway :slightly_smiling_face:

PR #58, opened 2021-12-13
[14.0][IMP] product_pack: Add tests

rousseldenis:


pedrobaeza:
Tests are always welcome, but I'm not sure this qualifies for being co-author of the module.

rousseldenis:
> Tests are always welcome, but I'm not sure this qualifies for being co-author of the module.

I was sure you answer that. But tests are part of development. Quite sure you did same on other PRs elsewhere...

pedrobaeza:
It doesn't matter if the code is test or other thing. I'm talking about the amount of the contribution respecting the whole module.

rousseldenis:
> It doesn't matter if the code is test or other thing. I'm talking about the amount of the contribution respecting the whole module.

How do you objectivate the amount of contribution ?

pedrobaeza:
Indeed it's not easy, but several metrics can be:

- Amount of lines. It's not too much as this module is the base.
- Amount of invested time. I can assure you we have spent much more than you on these modules.

rousseldenis:
> * Amount of invested time. I can assure you we have spent much more than you on these modules.

How can you know?

pedrobaeza:
> How can you know?

Seeing this PR... but you can tell me.

product_pack/tests/test_product_pack.py
```diff
@@ -0,0 +1,104 @@
...
+            "product_pack.product_pack_cpu_detailed_components"
+        ).product_tmpl_id
+        pack.pack_modifiable = True
+        with Form(pack) as pack_form:
+            pack_form.pack_type = "non_detailed"
+        self.assertFalse(pack_form.pack_modifiable)
+
+
+class ProductPackTest(TestProductPack, SavepointCase):
+    """ Test Launch """
```
lmignon: @rousseldenis Why do you add this class and not inherit from SavepointCase in TestProductPack? It's useless.

lmignon:
@pedrobaeza Writing tests is an essential element in the quality of a module. And the mere fact of writing them seems to me to be reason enough to be called a co-author. Ideally a new module should come with tests https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#tests

pedrobaeza:
The real use case tests are in sale_product_pack.

lmignon:
This one should also be tested on its own.

rousseldenis:
> The real use case tests are in sale_product_pack.

There is no 'real' use case tests. Tests are always contextually dependent of the module. In fact, sale_product_pack tests are done for ... 'sale' depending on the code added there. But you need to ensure that the base module behaviour is kept safe with its own tests.

pedrobaeza:
I know, but it's for the said argument of not having tests. Most of the things are already tested there, like prices. Anyways, as said, tests are always welcome, but my point still continues the same: you should annotate yourself as contributor, not co-author.

And please, don't use demo data on tests.

rousseldenis:
> And please, don't use demo data on tests.

Why ????? They are defined in this module...

product_pack/tests/test_product_pack.py
rousseldenis: @lmignon done

lmignon (approved):
LGTM (Code review only)

pedrobaeza:
> > And please, don't use demo data on tests.
> 
> Why ????? They are defined in this module...

This is something Odoo has finally changed the approach. Using demo data make that on future migrations, tests may fail not because the features itself, but because demo data have changed for whatever reason. It's one of the main principles we use when designing tests: don't depend on anything external (being demo data, default values, etc).

rousseldenis:
> This is something Odoo has finally changed the approach. Using demo data make that on future migrations, tests may fail not because the features itself, but because demo data have changed for whatever reason. It's one of the main principles we use when designing tests: don't depend on anything external (being demo data, default values, etc).

I know that.

What I said is I don't depend on Odoo demo data, but those defined here.

Could we go further as I removed author modifications ?

pedrobaeza (approved):
Still relying on your own demo data makes that changing such data in future improvements or migrations (adapting them to upstream changes), may lead to test failing, but I'm not going to insist more.

rousseldenis:
> Still relying on your own demo data makes that changing such data in future improvements or migrations (adapting them to upstream changes), may lead to test failing, but I'm not going to insist more.

Ok, I see what you want to say.

A better approach should be building test common to define test data. Is that you wanted to say ?

pedrobaeza:
Yes, it can be on your setup method, or better to have a separate class for being reused in other modules/tests (you have already done such pre-work creating 2 classes). An example I have is:

- Common class for all the Spanish fiscal reports: https://github.com/OCA/l10n-spain/blob/14.0/l10n_es_aeat/tests/test_l10n_es_aeat_mod_base.py
- It's used in VAT taxes report here for also defining another base for this model extensions: https://github.com/OCA/l10n-spain/blob/14.0/l10n_es_aeat_mod303/tests/test_l10n_es_aeat_mod303.py#L15
- And finally used for concrete tests: https://github.com/OCA/l10n-spain/blob/28d5a9611e7a490ca7c1d30fe36d640e57c73fc9/l10n_es_aeat_mod303/tests/test_l10n_es_aeat_mod303.py#L441
- And another module that uses such second generic test class for only adding the needed bits: https://github.com/OCA/l10n-spain/blob/14.0/l10n_es_aeat_mod303_oss/tests/test_l10n_es_aeat_mod303_oss.py

pedrobaeza:
It seems npm is not available. Let's try again:

/ocabot merge patch

pedrobaeza:
Still failing. We have to wait a bit to see if it's resolved.

PR #60, opened 2021-12-16
[14.0][IMP] product_pack: Change module to application

### LINKED ISSUE #59: Add product pack app
I would like your advice on this.

I want to make a 'product pack' app, but there are several approaches.

- Either we can define a product_pack_base that allows to define menus, maybe access rights, ... and base other modules on top of this
- Define product_pack module as 'application' and do stuff from above.
- Define a new module that aggregates the needed ones to build the app.

Thanks for your feedback

rousseldenis:
- [x] Add icon for application
- [x] Add user and manager groups
- [x] Add first action menu
- [x] ~~Modify access rules~~
- [x] ~~Set visibility for fields and tab~~
- [x] ~~Add record rules to see packs or not~~
- [x] ~~Add index on 'pack_ok' field~~


rousseldenis:
@pedrobaeza

product_pack/models/product_template.py
```diff
@@ -32,6 +32,7 @@ class ProductTemplate(models.Model):
     pack_ok = fields.Boolean(
         "Is Pack?",
         help="Is a Product Pack?",
+        index=True,
```
pedrobaeza: An index over a boolean field is not adding value.

product_pack/security/ir.model.access.csv
```diff
@@ -1,3 +1,3 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
-access_product_pack_line_user,product.pack.line,model_product_pack_line,base.group_user,1,0,0,0
-access_product_pack_line_system,product.pack.line,model_product_pack_line,base.group_system,1,1,1,1
```
pedrobaeza: This is what I meant of not changing current permissions. This mimics base `product` permissions (https://github.com/odoo/odoo/blob/14.0/addons/product/security/ir.model.access.csv#L3), and it's reasonable to keep them the same. You can add the extra 1,1,1,1 one for the product pack responsible though (the XML-ID is not correct right now also).

product_pack/security/security_rules.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2021 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+    <!-- Rules -->
+    <record id="product_pack_rule_global" model="ir.rule">
+        <field name="name">product.template.pack.user</field>
+        <field name="model_id" ref="product.model_product_template" />
+        <field name="domain_force">[('pack_ok', '=', False)]</field>
```
pedrobaeza: This record rule is going to cause a lot of troubles. For example, for e-commerce. Why do you consider it is needed?

product_pack/views/product_template_views.xml
```diff
@@ -9,7 +9,7 @@
         <field name="inherit_id" ref="product.product_template_form_view" />
         <field name="arch" type="xml">
             <div name="options" position="inside">
-                <div>
+                <div groups="product_pack.group_product_pack_user">
```
pedrobaeza: Groups should be at view level, not field level. This way, the access problems may arise.

pedrobaeza (changes requested):

product_pack/views/product_template_views.xml
rousseldenis: I don't agree. As for analytic for example, it allows to 'enable' a feature.

So, if user has access to packs, this div should be visible.

pedrobaeza: That's not what I'm saying. I'm talking only at technical level.

product_pack/security/security_rules.xml
rousseldenis: Indeed, maybe a little bit too much

product_pack/views/product_template_views.xml
rousseldenis: So ?

pedrobaeza: I mean that if you protect by record rules some related (m2o, m2m) fields, putting the group that grants the access to them at other level that is not the view one, you will get a permission error when accessed by users without such group.

rousseldenis: But aren't ir.rules evaluated as sudo ?

rousseldenis:
@pedrobaeza Changes done

rvalyi (approved):

rousseldenis:
@pedrobaeza Could we move forward on this ?

product_pack/views/product_template_views.xml
pedrobaeza: That's not what I mean. The question is that a field is included in the view definition, it's fetched, and then if you don't have access to target model, there will be an error. The only way to avoid that is to protect the entire view with the group, not only a part of it.

rousseldenis:
@pedrobaeza

product_pack/views/product_template_views.xml
rousseldenis: Done

PR #61, opened 2021-12-16
[14.0][IMP] product_pack: Change pack type to display type

rousseldenis:
Easy one as it is not a pack type.


pedrobaeza (approved):

rousseldenis:
Mmmh nearly, it displays warning with sale module (duplicate names).

Should be 'Pack Display Type' I think

pedrobaeza:
OK

rvalyi (approved):

PR #69, opened 2022-01-27
[13.0] [FIX] website_sale_product_pack: Fixed duplicated reward line image in cart summary

jinye-domatix:
For reward lines the image of the product in the cart summary appears two times.

### DIFF
--- a/website_sale_product_pack/views/templates.xml
+++ b/website_sale_product_pack/views/templates.xml
@@ -82,7 +82,7 @@
             expr="//span[@t-field='line.product_id.image_128']"
             position="attributes"
         >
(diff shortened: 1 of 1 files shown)


pedrobaeza:
You can't do this as the module doesn't depend on `sale_coupon`. You have to fix this in your custom or create a glue module.

jinye-domatix:
Okay I understand. I close the pull request

jinye-domatix:
Hi @pedrobaeza, if I create a glue module, should I do the PR to this repository?

pedrobaeza:
Yeah, it can be here or in sale-promotion, but I think it suits better here.

PR #77, opened 2022-03-16
[14.0][IMP] sale_product_pack: Several improvements

rousseldenis:
- [x] Add field for do not expand
- [x] Remove searches and use advantages of prefetch
- [x] Move sequence fill in to prepare values function
- [x] Add test function for create multi


rousseldenis:
Wanted also to do a real create multi.

But due to orm limitation, it is not possible (problem with sequencing as ids are created for main records before one2many ones even if passed in (0, 0, {}) => pack_child_line_ids)

cvinh (approved):
LGTM

pedrobaeza (approved):

PR #83, opened 2022-06-08
[13.0][FIX] price is shown in view instead of list_price (which the original view shows)

alvaro-domatix:
We detected that not the correct price was being shown in the website and we realized the inherited view replace the field list_price for the field price, which is not correct.


alvaro-domatix:
Can you review? @chienandalu

PR #85, opened 2022-06-08
[14.0][MIG] website_sale_product_pack

alvaro-domatix:


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

alvaro-domatix:
> Thanks for the contribution.
> 
> Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.
> 
> If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

Thank you. I've done it properly now but tests doesn't run. I've tried to run tests of product_pack module in my local instances and doesn't run either. Can I have some help about this? Looking at the Travis log, what am I doing wrong?

pedrobaeza:
Tests are executed, but there are errors on them. You have to fix them.

gregory-moka:
Few things are missing : 
**On website search field, the product price is not taking the pack setting to display the right price**

image

**Once you delete a detailed pack from your cart, the products defined in the pack are always displayed**

_Before :_ 
image

_After :_ 
image

**The amount of the detailed and price displayed pack is computed instead of having only the pack price**

image

**For ignored price pack, once you want to proceed to checkout, the amount change on the confirmation process :** 

_Before :_ 
image
 
_After :_ 
image

gregory-moka (changes requested):
The cart view is pretty nice, and the list of product included into the pack is cleverly presented.
But price computation is not working.M aybe something is missing on my setting, but I don't know what

PR #81, opened 2022-06-02
[15.0][MIG] sale_product_pack: Migration to 15.0

nicomacr:

PR #90, opened 2022-06-28
[15.0][MIG] sale_product_pack: Migration to 15.0 

AMhadhbi:


pedrobaeza:
Please split the changes for the migration itself (like manifest version bumping) in a separate commit as the migration commit states: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

pedrobaeza:
It was previously #81 and as this one is still not fulfilled, I close it in favor of the other.

PR #94, opened 2022-07-05
[13.0][FIX] website_sale_product_pack: Resilient tests

pedrobaeza:
Create and select a specific pricelist for avoiding problems in integrated environments where the default pricelist currency has been changed.

@Tecnativa


cesar-tecnativa (approved):

pedrobaeza:
Not enough, so resetting to draft

chienandalu (approved):
It was necessary to add the default pricelist to the portal user, which is the one used in the tours. I added the commit with the change, although you can squash it with yours.

PR #96, opened 2022-07-08
[IMP] stock_product_pack: incoming/outgoing stock

bruno-zanotti:
Before this commit, the module do not compute the incoming and outgoing pack stock.
Now is done following this logic:

- _Stock_on_hand_ = The number of packs that can be assembled with the stock components on hand.
- _Stock_on_hand_inc_ = The number of packs that can be assembled with the stock components on hand plus the incoming stock.
- Stock_on_hand_out = The number of packs that can be assembled with the stock components on hand minus the outgoing stock.

Then:

**Stock incoming** = _Stock_on_hand_inc_ - _Stock_on_hand_
This is the number of extra packs that we will be able to assemble if the incoming stock of the components arrives.

**Stock outgoing** = _Stock_on_hand_ + _Stock_on_hand_out_
This is the number of packs that we will NOT be able to assemble if the outgoing stock of the components is delivered

**Stock virtual available** = available + incoming - outgoing
Note that this is an improvement because now we are following the same logic for virtual stock as Odoo in normal products.

### DIFF
--- a/stock_product_pack/models/product_product.py
+++ b/stock_product_pack/models/product_product.py
@@ -17,39 +17,72 @@ class ProductProduct(models.Model):
             lot_id, owner_id, package_id, from_date=from_date, to_date=to_date
         )
         packs = self.filtered("pack_ok")
+
+        def pack_quantity(components):
+            """ Returns the number of packs that can be assembled with the stock of
+                the components.
+                :param components: List of tuples (stock, quantity required for pack)
+            """
+            return (
+                components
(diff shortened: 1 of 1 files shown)


nicomacr (approved):
Technical review LGTM

bruno-zanotti:
@ernestotejeda @pedrobaeza could you check this PR? The idea is to calculate the outgoing/incoming stock for product packs

PR #97, opened 2022-07-11
[13.0][FIX] product_pack: Compute directly the price on pricelists

pedrobaeza:
Steps to reproduce the problem:

- Create a new pack with "Pack component price" = "Totalized in main product"
- Put some components on it.
- Create a sales order with such pack.
- The price unit got in the sales order line is not the sum of the components.

That's because the computations done after calling `price_compute` in the pricelist engine modify the final returned price, so we don't call super on such cases and compute it directly.

@Tecnativa TT37977

### DIFF
--- a/product_pack/models/__init__.py
+++ b/product_pack/models/__init__.py
@@ -1,5 +1,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from . import product_pack_line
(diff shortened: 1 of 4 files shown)


cesar-tecnativa (approved):

nicomacr (approved):
LGTM

nicomacr:
@jjscarafia could you take a look at this?

PR #100, opened 2022-09-09
[FIX] product_pack: compute price pack detailed totalized

bruno-zanotti:
For sale orders that have a Pack (detailed and totalized in main product) and the pricelist has an operation (i.e. list_price + 10%), the result was wrong, adding +10% twice.
In this PR, we only calculate the price with pricelist for the components of the pack and not for the pack

### DIFF
--- a/product_pack/models/__init__.py
+++ b/product_pack/models/__init__.py
@@ -3,3 +3,4 @@
 from . import product_pack_line
 from . import product_product
 from . import product_template
+from . import product_pricelist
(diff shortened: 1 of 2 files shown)


bruno-zanotti:
Hi @ernestotejeda could you review this PR? There is an issue when a product pack price is computed.

bruno-zanotti:
Merged in https://github.com/OCA/product-pack/pull/97

PR #105, opened 2022-10-21
[WIP][FIX] product_pricelist: pricelist based on pricelist with different currency

augusto-weiss:
Steps to replicate:
* Activate multi-currency
* Activate multi-pricelist (Advanced price rules)
* Create a pricelist (EUR currency) based on "other pricelist" (could be Public Pricelist which is set in USD)
* Create a sale order with the EUR pricelist and add a Product Pack:
The price should be in EUR but is computed in USD.

### DIFF
--- a/product_pack/models/__init__.py
+++ b/product_pack/models/__init__.py
@@ -1,5 +1,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from . import product_pack_line
+from . import product_pricelist
 from . import product_product
 from . import product_template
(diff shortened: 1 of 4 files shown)


augusto-weiss:
Hi @pedrobaeza, we detect some issues with product packs. Also found your PR ( #97 ) so we tested it and added some features to this PR. Could you check it and give us a review ??
Thank you! Regards.

cc @nicomacr

pedrobaeza:
If the other PR is correct, please approve it and let's merge it.

About this one, please create a test that recreates the problem you are having, so that it serves as showcase and also as a regression check.

pedrobaeza:
Closing as no answer.

PR #113, opened 2022-12-09
[15.0][FIX] product_pack: Compute directly the price on pricelists

pedrobaeza:
Forward-port of #97 #106

Steps to reproduce the problem:

- Create a new pack with "Pack component price" = "Totalized in main product"
- Put some components on it.
- Create a sales order with such pack.
- The price unit got in the sales order line is not the sum of the components.

That's because the computations done after calling `price_compute` in the pricelist engine modify the final returned price, so we don't call super on such cases and compute it directly.

@Tecnativa TT37977


victoralmau (approved):

rafaelbn (approved):
👍

PR #104, opened 2022-10-19
[16.0] [MIG] sale_product_pack: Migration to 16.0

JulyChacko1601:

PR #121, opened 2023-02-22
[16.0][MIG] sale_product_pack: Migration to 16.0

FrancoMaxime:
This pull request is based on the closed pull request (https://github.com/OCA/product-pack/pull/104).


sale_product_pack/models/product_pack_line.py
```diff
@@ -24,10 +24,11 @@ def get_sale_order_line_vals(self, line, order):
...
         }
         sol = line.new(line_vals)
-        sol.product_id_change()
+        sol._onchange_product_id_warning()
         sol.product_uom_qty = quantity
-        sol.product_uom_change()
-        sol._onchange_discount()
+        sol._compute_product_uom_qty()
+        sol._compute_product_uom()
+        sol._compute_discount()
```
rven: Is the calling of these functions really needed?

Why is the function **_onchange_product_id_warning** needed to be called?

**_compute_product_uom_qty** is strange to call it because you set **product_uom_qty**

**_compute_uom** does not use **product_uom_qty** in its calculation

**_compute_discount** depends on **product_uom_qty**

If I remove them all, the test scenario's succeed.
So if they are needed I suppose a test scenario should be created.

rven (changes requested):

sale_product_pack/models/product_pack_line.py
FrancoMaxime: To be honest, i picked up the changes from July after she closed her pr. I didn't think further. I will check that

FrancoMaxime: Hi @rven, 
You are right the _compute are not needed. 
The _onchange_product_id_warning was previously inside the product_id_change. To avoid losing this functionality, I think we should call the onchange method.

rven (approved):

sanrav:
I've tested 4 different pack configurations. Here are my results : 

**Testing not OK** : 

* **Pack type = Non detailed** : 
   * Pack components not displayed in the SO and sales price of the pack on SO line = total price of components => **OK**
   * In V15, the field Sales price on the pack configuration displays automatically the total price of the components but not in V16. Maybe, it's better to keep the same functionality as in V15 and calculate automatically the sales price on the product.template
   * The price of the pack on the SO line is not updated when I choose a pricelist containing, for instance, a discount of 10% for this pack. I click on "Update prices" next to the pricelist on the SO but the price on the line doesn't change. Even if I choose first the pricelist before choosing my pack on the line, the price of the pack doesn't take into account the pricelist. I think we should take into account the pricelist for non detailed packs.
* **Pack type = Detailed and Price = Totalized in main product** : 
   * Pack components displayed in the SO and sales price of the pack = total price of components (prices of the components not displayed in the SO) => **OK**
   * Same issues as above regarding the prices but in addition, when I click on "Update price" after changing the pricelist on the SO, ODOO updates the sales price of the components and displays the prices on the SO lines. It should not display the prices of the components.
* **Pack type = Detailed and Price = Ignored** : 
   * Pack components displayed in the SO and sales price of the pack = the one defined on its product configuration (not linked with component prices). Prices of the components not displayed in the SO => **OK**
   * When I click on "Update price" after changing the pricelist on the SO, ODOO updates the sales price of the components and displays the prices on the SO lines. It should not display the prices of the components.

**Testing OK** : 

* **Pack type = Detailed and Price = Detailed per component** : 
   * Pack components displayed in the SO and sales price of the pack and of the components displayed on SO lines. Price of the pack are the one defined on its product configuration (not linked with component prices) => **OK**
   * Prices on SO lines properly updated with pricelist => **OK**

FrancoMaxime:
Two new commits have been added. The first allowing product_pack to use pricelist, and the second allows you to correctly calculate the prices. Unit test have been updated to check if the computation of product_pack price after pricelist update works. (for each product_pack type.)

sanrav (approved):
Functional testing OK

jjscarafia:
Hi @FrancoMaxime ! We're already using this PR for some customers, working ok except from the bug and fix proposed here https://github.com/acsone/product-pack/pull/1
Thanks!

FrancoMaxime:
Hi @jjscarafia, i made a little comment on it.

sale_product_pack/models/sale_order_line.py
```diff
@@ -73,11 +73,20 @@ def expand_pack_line(self, write=False):
...
-        record = super().create(vals)
-        record.expand_pack_line()
-        return record
+    @api.model_create_multi
+    def create(self, vals_list):
+        new_vals = []
+        res = self.browse()
+        for elem in vals_list:
+            product = self.env["product.product"].browse(elem["product_id"])
+            if product.pack_ok and product.pack_type == "detailed":
```
augusto-weiss: if product.pack_ok and product.pack_type != "non_detailed":

@FrancoMaxime Hi Franco again! Could you update this line? Because its necessary to be able to call the method expand_pack_line when the module sale_product_pack_assisted is installed... and there the product pack type is non_detailed_assisted.

Thanks!!

FrancoMaxime: @augusto-weiss Hi, i read, your module and mine. I think that's a bad idea to add a method who returns all pack_type for the create because we can't use it in the other condition. You should overwrite the create, split the records in self into two categories, the first one for the super() call and the second one, for the work you want for your product pack. What do you think about this?

augusto-weiss: https://github.com/acsone/product-pack/pull/2

bealdav:
FInally mergeable ?

JrAdhoc:
@FrancoMaxime @pedrobaeza 
Is this ready for merging??

FrancoMaxime:
@JrAdhoc Hi, for me, it is :)

nicolascol:
Hi! functional review LGTM +1
Can you merge it?
@rousseldenis

rousseldenis:
@FrancoMaxime Could you rearrange commits (essentially delete merge ones) ?

FrancoMaxime:
@rousseldenis  I did it.

alvaro-domatix (approved):

PR #126, opened 2023-02-27
[FIX] product_pack,sale_product_pack: change ambiguous variable for lambda functions

ernesto-garcia-tecnativa:


.pre-commit-config.yaml
```diff
@@ -33,14 +33,14 @@ repos:
         language: fail
         files: '[a-zA-Z0-9_]*/i18n/en\.po$'
   - repo: https://github.com/oca/maintainer-tools
-    rev: dfba427ba03900b69e0a7f2c65890dc48921d36a
+    rev: e3b46970af62cbbf99f0b45a6675eb4cae744d6d
```
pedrobaeza, line 36: Why this file has been modified without copier update?

PR #129, opened 2023-03-30
[14.0][IMP][sale_product_pack] Add prices in Product pack lines

cvinh:
This can be helpfull to define sale price for the product pack
We added 2 computed fields total standard price and total public price in product pack lines to help the user to define the pack sale price


elvise:
@cvinh can you rebase ?

PR #133, opened 2023-05-15
[16.0][FIX] product_pack: doing get_price more like price_compute method to avoid errors

augusto-weiss:


augusto-weiss:
@FrancoMaxime 
Hi Franco, I made this PR based on yours (#121) to fix problems getting price when the pricelist is based on Cost (standard_price).
So, the most important change is the get_price method which I try to make seem more like the price_compute method.

To reproduce the error:
Create a pricelist based on cost
Create a product pack and add some components
Add the pack to a new sale order with the pricelist based on cost
The price unit is the sum of components list price and not the sum of cost price

How looks this fix?

product_pack/models/product_pack_line.py
```diff
@@ -54,6 +54,14 @@ def _check_recursion(self):
...
                 pack_lines = pack_lines.mapped("product_id.pack_line_ids")
 
-    def get_price(self):
+    def get_price(self, price_type, currency=False, company=False, date=False):
         self.ensure_one()
+        if price_type == "standard_price":
+            price_currency = self.product_id.cost_currency_id
+            company = company or self.env.company
+            date = date or fields.Date.context_today(self)
+            price = self.product_id["standard_price"]
```
FrancoMaxime:
```suggestion
            price = self.product_id.standard_price
```

FrancoMaxime (changes requested):
I tested it functionally and its working. I think this PR should wait the merge of #121. Nice work!

product_pack/models/product_pack_line.py
augusto-weiss: Yes, totally! thats why i made a new PR! Thanks for reviewed it

FrancoMaxime (approved):

augusto-weiss:
@FrancoMaxime Sure!! thats why i made a new PR!! thanks a lot

augusto-weiss:
Hi @FrancoMaxime i saw your PR was merged, good job and thanks for contributing!!

I resume this PR, could you review it ?

FrancoMaxime (approved):
LGTM code review and functional test. Maybe we could add a unit test to make sure the problem doesn't come back.

augusto-weiss:
Thanks @FrancoMaxime  i added a test case !

augusto-weiss:
Hi @rousseldenis and @pedrobaeza , Could you too review this PR? Thanks a lot

sale_product_pack/tests/test_sale_product_pack.py
```diff
@@ -169,6 +186,14 @@ def test_create_totalized_price_order_line(self):
             (self.sale_order.order_line - line).mapped("price_subtotal"), [0, 0, 0]
         )
```
pedrobaeza: Please don't put empty lines inside a method.

pedrobaeza (commented):
Can you please expose the steps to reproduce a problem which this code is avoiding?

sale_product_pack/tests/test_sale_product_pack.py
rousseldenis: > Can you please expose the steps to reproduce a problem which this code is avoiding?

@pedrobaeza Exposed there : https://github.com/OCA/product-pack/pull/133#issuecomment-1548115985

augusto-weiss:
Hi @pedrobaeza and @rousseldenis thanks for answer!
I updated the changes.

The issue:
problems getting price when the pricelist is based on Cost (standard_price).

To reproduce the error:
Create a pricelist based on cost
Create a product pack and add some components
Add the pack to a new sale order with the pricelist based on cost
The price unit is the sum of components list price and not the sum of cost price

pedrobaeza (commented):
I now understand the problem you are mentioning, but what I don't get is why don't we just rely on Odoo itself for getting the proper sales price instead of redoing the pricelist engine.

rousseldenis:
> I now understand the problem you are mentioning, but what I don't get is why don't we just rely on Odoo itself for getting the proper sales price instead of redoing the pricelist engine.

@augusto-weiss I think @pedrobaeza is right as going up the stack make me thing we should use quite the same process done in sale : 

https://github.com/odoo/odoo/blob/16.0/addons/sale/models/sale_order_line.py#L437

What do you think of ?

@FrancoMaxime

augusto-weiss:
That's interesting!! But at the end of all, the method _get_display_price (defined in sale order line) call the method price_compute, which we are redefining in product_pack, and before this PR didn't consider the pricelist type (price_type in odoo: https://github.com/odoo/odoo/blob/16.0/addons/product/models/product_product.py#L609).

So, working at sale order line level, we need to redefine some workflow getting the pack price!

Right ? or I am confused ?
@pedrobaeza @rousseldenis  @FrancoMaxime

pedrobaeza:
But IMO we should call `_get_display_price` for each component on the `price_compute` overwrite, which is the aim of this.

FrancoMaxime:
Hi,
@augusto-weiss @pedrobaeza  and @rousseldenis  I agree with you when you say the compute method for the price should approach the standard. But It will be hard for a product.pack.line to use a method inside a sale.order.line that needs information from a sale.order. But maybe i'm missing some information to see how to that without creating a sale.order with the new() method

pedrobaeza:
Hi @FrancoMaxime, you can extract a bit of the code from `sale` to get it. Something like:

```python
order = pack_line.order_id
order_date = order.date_order or fields.Date.today()
product = pack_line.product_id
qty = pack_line.product_uom_qty or 1.0
uom = pack_line.product_uom or product.uom_id
pricelist_rule = order.pricelist_id._get_product_rule(product, qty, uom=uom, date=order_date)
price = pricelist_rule._compute_price(product, qty, uom, order_date, currency=order.currency_id)
```

Although you can also instantiate a virtual sale.order.line (with `new`) and get the price.

nicolascol:
Hi guys!
I tested this PR and the behavior was as expected. Solves the problem about the price calculation when the pricelist it's based on cost

bruno-zanotti (approved):
Technical review LGTM. Odoo made lots of changes in prices and pricelists in 16, so probably we will need more changes in product_pack, but at least this is an improvement of what we had 👍🏼

pedrobaeza (changes requested):
Please check my previous comments

bruno-zanotti:
Hey @pedrobaeza,

I've read the comments and I agree with you that redoing the pricelist engine might not be the best solution. However, I'm still not sure what the best approach is, maybe I’m missing something, but _get_display_price it only defined for sale.order.line, so to start we should move this to sale_product_pack.. but then what happen with other models that ends calling price_compute? for example [here](https://github.com/odoo/odoo/blob/5cc72bb6d6f9a04ec32e47d02ebfaf8e12aba559/addons/account/models/account_analytic_line.py#L74) or the method [_compute_price](https://github.com/odoo/odoo/blob/3619f77c73092e800aab10f0df09ca5775545da3/addons/product/models/product_pricelist_item.py#L351)

I suggested trying with the method [_get_contextual_price](https://github.com/odoo/odoo/blob/561eba8f2adece4c21ea46c1bd050779ea3dc676/addons/product/models/product_template.py#L1309) but we don't get any better solution..

pedrobaeza:
Why my proposal in https://github.com/OCA/product-pack/pull/133#issuecomment-1598484645 is not valid?

bruno-zanotti:
Because we don't have an order yet... and making a virtual sale.order.line wouldn't work because of what I said before.
I think the method get_price is more generic, that's why I think we should use _get_contextual_price..

pedrobaeza:
You should have already the date and the pricelist, so it's not involving to create a virtual SO line. Please check again.

augusto-weiss:
Hi @pedrobaeza i implemented some of your recommendations! Please take a look
First, I reverted the changes and then implemented the code but in sale_product_pack, the problem is that these changes don't work for e-commerce, maybe need to be implemented in website_product_pack also ?

pedrobaeza:
Can you please put CI green?

bruno-zanotti:
Hi @augusto-weiss, I made a small fix in a new commit to avoid computing the pack price when the pack is detailed by components, please review and squash if you think it's ok.

To replicate the bug just add a pack detailed per component to a sale order and check the prices

PR #134, opened 2023-06-06
[16.0][FIX] stock_product_pack Use already calculated qtys

alvaro-domatix:
There are some cases where it may occur that the pack quantities and the components are attempted to be calculated at the same time, resulting in a logical error. When trying to calculate the pack quantities based on their components and these components being calculated at that very moment, the components return quantities equal to 0 and the outcome is that the calculation of the pack quantities is not consistent.

This happens, for instance, in remote cases where on a delivery note all products (pack + components) are attempted to be added, or in more common cases related to other modules from the same repository that necessitate the creation of a module to solve it.

With this fix, we solve a consistency error that may occur in this module independently and in future and parallel modules.

### DIFF
--- a/stock_product_pack/tests/test_stock_product_pack.py
+++ b/stock_product_pack/tests/test_stock_product_pack.py
@@ -175,6 +175,10 @@ class TestSaleProductPack(TransactionCase):
         wizard.process()
         self.assertEqual(self.pack_dc.virtual_available, 5)
         self.assertEqual(self.pack_dc.qty_available, 5)
+        all_product_ids = self.env["product.product"]
+        all_product_ids |= self.pack_dc
+        all_product_ids |= components
+        all_product_ids._compute_quantities()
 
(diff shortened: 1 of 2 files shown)


stock_product_pack/tests/test_stock_product_pack.py
```diff
@@ -175,6 +175,10 @@ def test_compute_quantities_dict(self):
         wizard.process()
         self.assertEqual(self.pack_dc.virtual_available, 5)
         self.assertEqual(self.pack_dc.qty_available, 5)
+        all_product_ids = self.env["product.product"]
+        all_product_ids |= self.pack_dc
+        all_product_ids |= components
+        all_product_ids._compute_quantities()
```
pedrobaeza, line 181: Why do you need to call manually the compute? This means a problem elsewhere.

pedrobaeza (commented):
Can you please create a test case showcasing the previous problem?

PR #135, opened 2023-06-07
[16.0] [MIG] website_sale_product_pack: Migration to 16.0

alvaro-domatix:


alvaro-domatix:
Hi, @ernestotejeda @pedrobaeza !

I would appreciate if you could review this and #134.

pedrobaeza:
/ocabot migration website_sale_product_pack

I'm not the most suitable for reviewing this. @chienandalu if you may?

liebana:
Functional review, LGTM.

As an small improvement (maybe not here), if Pack Display Type is "Detailed" once we add the pack to the cart I think the number of items should be the sum of the child lines instead of the number of packs.

image

liebana (approved):

augusto-weiss:
Hi @alvaro-domatix i'm testing this and looks good! But, I notice a problem with packs detailed per components. On website the price is computed in a wrong way. Do you have any idea about that ?
I am working on that but didn't find a good fix yet! 
The problem is that somewhere the context "whole_price_pack" is lost

augusto-weiss:
Hi @alvaro-domatix, @pedrobaeza and @chienandalu 
I added this pr #144 to fix the problem I mentioned before, could you review it?
Thanks a lot!!

sbiosca-s73 (approved):

website_sale_product_pack/__manifest__.py
```diff
@@ -4,10 +4,15 @@
     "name": "Website Sale Product Pack",
     "category": "E-Commerce",
     "summary": "Compatibility module of product pack with e-commerce",
-    "version": "13.0.1.0.2",
+    "version": "16.0.1.0.3",
```
chienandalu, line 7: You just need to bump the major version and reset the minor digits :)
```suggestion
    "version": "16.0.1.0.0",
```

website_sale_product_pack/models/product_template.py
```diff
@@ -33,8 +40,54 @@ def check_website_published(self):
...
+                        "product_name": rec.name,
+                        "pack_parents": ", ".join(published.mapped("name")),
+                    }
                 )
+
+    # Neccessary for the website_sale_product_pack module because the price in /shop
+    # is calculated by the product.template.price_compute method
+    def price_compute(
+        self, price_type, uom=False, currency=False, company=False, date=False
+    ):
```
chienandalu, line 56: Wouldn't this override belong to the base module? Isn't there a common method there we can use?

chienandalu (commented):
Thanks @alvaro-domatix :D

Some comments:

website_sale_product_pack/models/sale_order.py
```diff
@@ -0,0 +1,53 @@
...
+                and pack.pack_component_price in {"totalized", "ignored"}
+            )
+            return super(
+                SaleOrder,
+                self.with_context(
+                    pack_discount=line.discount,
+                    detailed_totalized_pack=detailed_totalized_pack,
+                ),
+            )._cart_update(product_id, line_id, add_qty, set_qty, **kwargs)
+        return super()._cart_update(product_id, line_id, add_qty, set_qty, **kwargs)
```
ioans73, line 29:
```suggestion
    def _cart_update(
        self, *args, **kwargs
    ):
        """We need to keep the discount defined on the components when checking out.
        Also when a line comes from a totalized pack, we should flag it to avoid
        changing it's price in a cart step."""
        line_id = kwargs.get("line_id")
        if line_id:
            line = self.env["sale.order.line"].browse(line_id)
            if line and line.pack_parent_line_id:
                pack = line.pack_parent_line_id.product_id
                detailed_totalized_pack = (
                    pack.pack_type == "detailed"
                    and pack.pack_component_price in {"totalized", "ignored"}
                )
                return super(
                    SaleOrder,
                    self.with_context(
                        pack_discount=line.discount,
                        detailed_totalized_pack=detailed_totalized_pack,
                    ),
                )._cart_update(*args, **kwargs)
        return super()._cart_update(*args, **kwargs)
```

@alvaro-domatix Odoo has recently made these [changes](https://github.com/odoo/odoo/commit/b121f8e1d3ca5b98e71e25710427e6210afe9ef6) and now I have an error caused by this module and that I have been able to solve in this way. Can you apply the changes? Or if you think it should be done another way...


```
  File "/odoo/odoo/addons/website_sale_delivery/models/sale_order.py", line 76, in _cart_update
    return super()._cart_update(*args, **kwargs)
  File "/oca/product-pack/website_sale_product_pack/models/sale_order.py", line 29, in _cart_update
    return super()._cart_update(product_id, line_id, add_qty, set_qty, **kwargs)
  File "/odoo/odoo/addons/website_sale_loyalty/models/sale_order.py", line 160, in _cart_update
    product_id, set_qty = kwargs['product_id'], kwargs.get('set_qty')
KeyError: 'product_id'
```

ioans73: ping @alvaro-domatix

lef-adhoc: LGTM @alvaro-domatix

pedrobaeza:
Superseded by #165

PR #137, opened 2023-07-07
[15.0][MIG] sale_stock_product_pack

JasminSForgeFlow:
Standard Migration

@ForgeFlow


GuillemCForgeFlow (commented):
Just as a suggestion, maybe it would be better to only show Integers in the delivered quantities for the lines with Pack products. See image below.

image

It may be more understandable to set the delivered quantity to 1 only when all pack products have been completely delivered.

WDYT? @pedrobaeza @ernestotejeda

pedrobaeza:
I think it's better to see the fractions delivered as it is.

GuillemCForgeFlow (approved):
LGTM! 👍🏽

MiquelRForgeFlow (approved):

PR #141, opened 2023-08-04
[16.0][FIX] sale_product_pack: keep order lines in order

augusto-weiss:
The issue:
When add diferent sale orders, packs and their components are added first, and then the other order lines

To reproduce the error:
New sale order
(1) Add a sale order line 
(2) Add a sale order line (pack)
Save the sale order

Behavior before this PR:
The sale order will show:
(2) The pack sale order line and components first
(1) The other sale order line

Behavior after this PR:
Hold the order of sale order lines 1 and 2

[Untitled_ Aug 4, 2023 3_14 PM.webm](https://github.com/OCA/product-pack/assets/90717087/df10d2f5-13cd-4692-93ce-2375a4fae42c)

### DIFF
--- a/sale_product_pack/models/sale_order_line.py
+++ b/sale_product_pack/models/sale_order_line.py
@@ -75,7 +75,6 @@ class SaleOrderLine(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        new_vals = []
         res = self.browse()
         for elem in vals_list:
             product = self.env["product.product"].browse(elem.get("product_id"))
@@ -84,8 +83,7 @@ class SaleOrderLine(models.Model):
                 line.expand_pack_line()
                 res |= line
             else:
-                new_vals.append(elem)
-        res |= super().create(new_vals)
+                res |= super().create([elem])
         return res
 
     def write(self, vals):


sale_product_pack/models/sale_order_line.py
```diff
@@ -84,8 +84,7 @@ def create(self, vals_list):
                 line.expand_pack_line()
                 res |= line
             else:
-                new_vals.append(elem)
-        res |= super().create(new_vals)
+                res |= super().create([elem])
```
pedrobaeza, line 86: This is incorrect, as it's serializing the creation. If you want to keep order, play with other things like sequence, but I don't understand why they aren't properly ordered already. Please include steps to reproduce the problem.

pedrobaeza (changes requested):

sale_product_pack/models/sale_order_line.py
augusto-weiss: Hi! I added the steps to reproduce error and a video on description

pedrobaeza: The video is not shown, but as said, the solution for the problem shouldn't be to serialize the creation.

augusto-weiss: Hi @pedrobaeza I made some changes!

sale_product_pack/models/sale_order_line.py
```diff
@@ -71,13 +71,14 @@ def expand_pack_line(self, write=False):
...
-                self.create(vals_list)
+                return super().create(vals_list)
 
     @api.model_create_multi
     def create(self, vals_list):
         new_vals = []
         res = self.browse()
-        for elem in vals_list:
+        for idx, elem in enumerate(vals_list):
+            elem["sequence"] = idx
```
pedrobaeza: Why doing it here? Prepare the sequence on `expand_pack_line`, summing the main line sequence + an incremental number

augusto-weiss: All sale order lines have the same sequence (equal 10).
Editing the component lines also broke the sequence when you have some order lines added.

Example:
sale order line 1 (sequence = 10)
sale order line 2 (sequence = 10) -> pack detailled
sale order line 3 (sequence = 10)

Computed:
sale order line 1 (sequence = 10)
sale order line 2 (sequence = 10) -> pack detailed
   Component 1 (sequence = 11)
   Component 2 (sequence = 12)
sale order line 3 (sequence = 10)

Output result:
sale order line 1 (sequence = 10)
sale order line 2 (sequence = 10) -> pack detailled
sale order line 3 (sequence = 10)
   Component 1 (sequence = 11)
   Component 2 (sequence = 12)

augusto-weiss: Otherside, i must make some change, adding a context because every case have a different behaivour

pedrobaeza: This is still not correct, as you can make the create of the sales order lines on several batches, so you must check the resequence mechanism implemented in this version, and hook it to respect the relative pack lines orders.

PR #142, opened 2023-08-07
[16.0][ADD] purchase_product_pack

QuocDuong1306:
This module adds `Product Pack` functionality to purchase orders. You can choose a `Pack` in `purchase order lines` and see different behaviors depending on "Pack type" and "Pack component price" fields options selected on this `Pack`.

Depend on:
- [product_pack](https://github.com/OCA/product-pack/tree/16.0/product_pack) module


purchase_product_pack/models/purchase_order.py
```diff
@@ -0,0 +1,61 @@
...
+from odoo.exceptions import UserError
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    def copy(self, default=None):
+        purchase_copy = super().copy(default)
+        # we unlink pack lines that should not be copied
+        pack_copied_lines = purchase_copy.order_line.filtered(
```
simahawk: not a blocker but it would be better to exclude them from `copy_data` instead of creating and deleting them

purchase_product_pack/models/purchase_order.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        pack_copied_lines.unlink()
+        return purchase_copy
+
+    @api.onchange("order_line")
+    def check_pack_line_unlink(self):
+        """At least on embeded tree editable view odoo returns a recordset on
+        _origin.order_line only when lines are unlinked and this is exactly
+        what we need
+        """
```
simahawk: not a docstring explaining what the method does
```suggestion
    def check_pack_line_unlink(self):
        # At least on embeded tree editable view odoo returns a recordset on
        # _origin.order_line only when lines are unlinked and this is exactly
        # what we need
        """
```

purchase_product_pack/models/purchase_order.py
```diff
@@ -0,0 +1,61 @@
...
+        origin_line_ids = self._origin.order_line.ids
+        line_ids = self.order_line.ids
+        removed_line_ids = list(set(origin_line_ids) - set(line_ids))
```
simahawk: no need to convert to list

purchase_product_pack/models/purchase_order.py
```diff
@@ -0,0 +1,61 @@
...
+        ):
+            raise UserError(
+                _(
+                    "You cannot delete this line because is part of a pack in"
+                    " this purchase order. In order to delete this line you need to"
+                    " delete the pack itself"
+                )
+            )
+
+    # This function checks whether deleted_line (including subpack lines)
```
simahawk: this is a write method. I would move this check out to `_check_deleted_line` and then you can describe it w/ a proper docstring because this comment is not explaining what the method is doing ;)

purchase_product_pack/models/purchase_order_line.py
```diff
@@ -0,0 +1,211 @@
...
+        do_not_expand = self.env.context.get("update_prices") or self.env.context.get(
+            "update_pricelist", False
+        )
+        self.update(
+            {
+                "do_no_expand_pack_lines": do_not_expand,
+            }
+        )
+
+    def expand_pack_line(self, write=False):
```
simahawk, line 51: a docstring would be nice

purchase_product_pack/models/purchase_order_line.py
```diff
@@ -0,0 +1,211 @@
...
+        new_vals = []
+        res = self.browse()
+        for elem in vals_list:
+            product = self.env["product.product"].browse(elem.get("product_id"))
+            if product and product.pack_ok and product.pack_type != "non_detailed":
+                line = super().create([elem])
+                line.expand_pack_line()
+                res |= line
+            else:
+                new_vals.append(elem)
```
simahawk:
```suggestion
        prod_ids = [x["product_id"] for x in vals_list]
        products = self.env["product.product"].browse(prod_ids)
        for line_vals, product in zip(vals_list, products):
            if product and product.pack_ok and product.pack_type != "non_detailed":
                line = super().create([line_vals])
                line.expand_pack_line()
                res |= line
            else:
                new_vals.append(line_vals)
```

purchase_product_pack/models/product_product.py
```diff
@@ -0,0 +1,16 @@
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def cost_compute(self):
```
simahawk: judging from the docstring this method should be renamed to `pack_cost_compute`

purchase_product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,60 @@
...
+                    "> " * (line.pack_depth + 1), pol.product_id.name
+                ),
+            }
+        )
+        return vals
+
+    def get_cost(self):
+        """This function returns the cost of pack lines if they has seller or not"""
+        self.ensure_one()
+        if self._context.get("line", False):
```
simahawk: I don't get why we need to pass the line via ctx when the method can receive this record as an argument

simahawk: also, the line is used only for some parameters like date and uom that could be given as explicit arguments.
Thus, the method could be renamed `get_seller_cost`.
But I might be missing some implementation details of course...

purchase_product_pack/models/purchase_order_line.py
```diff
@@ -0,0 +1,211 @@
...
+    @api.depends("product_qty", "product_uom")
+    def _compute_price_unit_and_date_planned_and_name(self):
+        res = super()._compute_price_unit_and_date_planned_and_name()
+        # Re-calculate the price unit following options on product-pack
+        for line in self:
+            if not line.product_id or line.invoice_lines:
+                continue
+
+            params = {"order_id": line.order_id}
+            seller = line.product_id._select_seller(
```
simahawk, line 147: this is done in few places: can't we have a method to get the seller for the line?

purchase_product_pack/models/purchase_order_line.py
```diff
@@ -0,0 +1,211 @@
...
+                partner_id=line.partner_id,
+                quantity=line.product_qty,
+                date=line.order_id.date_order and line.order_id.date_order.date(),
+                uom_id=line.product_uom,
+                params=params,
+            )
+
+            prices = line.product_id.with_context(line=line).cost_compute()
```
simahawk: not sure we need this line in the ctx, see my other comment

purchase_product_pack/models/purchase_order_line.py
```diff
@@ -0,0 +1,211 @@
...
+            if not prices:
+                continue
+            else:
```
simahawk: no need for this else

purchase_product_pack/models/purchase_order_line.py
```diff
@@ -0,0 +1,211 @@
...
+                cost = prices[line.product_id.id]
+            # If not seller, use the standard price. It needs a proper currency conversion.
+            if not seller:
```
simahawk, line 161: this block should be moved out to its own method

simahawk: hmm I see this method has been partially copied from the odoo core one. Can you mention this and highlight the parts that have been modified?

simahawk (changes requested):

purchase_product_pack/models/purchase_order_line.py
QuocDuong1306: I think we can reuse `_select_seller` in products to get the seller

QuocDuong1306:
Thank @simahawk , i have updated. Could you have a look? Thanks

cyrilmanuel (approved):
not an expert but lgtm

purchase_product_pack/tests/test_purchase_product_pack.py
```diff
@@ -0,0 +1,289 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo.tests.common import TransactionCase
+
+
+class TestPurchaseProductPack(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.purchase_order = cls.env["purchase.order"].create(
```
simahawk, line 14:
```
from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
[...]
cls.env = cls.env(context=dict(cls.env.context, **DISABLED_MAIL_CONTEXT)
```

purchase_product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import models
+
+
+class ProductPack(models.Model):
```
simahawk:
```suggestion
class ProductPackLine(models.Model):
```

purchase_product_pack/models/product_pack_line.py
```diff
@@ -0,0 +1,59 @@
...
+                uom_id=line.product_uom,
+                params=params,
+            )
+            return (
+                pack_line_seller.price * self.quantity
+                if pack_line_seller
+                else self.product_id.standard_price * self.quantity
+            )
+        else:
+            return self.product_id.standard_price * self.quantity
```
simahawk:
```suggestion
        return self.product_id.standard_price * self.quantity
```

purchase_product_pack/models/purchase_order.py
```diff
@@ -0,0 +1,74 @@
...
```
simahawk: you can drop this comment

simahawk (approved):
pre-approving

cyrilmanuel:
@QuocDuong1306  ping :)

QuocDuong1306:
Hello, I have updated

cyrilmanuel (approved):

PR #146, opened 2023-09-06
[16.0][FIX] compute quantities context domain

augusto-weiss:


linuxivan:
No description, no info, why this? @augusto-weiss

PR #149, opened 2023-10-07
[REF] x_product_pack: pricing refactor 

bruno-zanotti:
### Pricing refactor
This PR aims to perform a thorough refactor of the **product_pack, sale_product_pack and website_sale_product_pack** modules. The primary goal is to enhance the **pricing** logic for product packs, ensuring consistency and improved functionality.

Key improvements / fixes:

1. **Refactor Pack Price Computation:** Removed the inheritance from price_compute in favor of a clearer and
    more comprehensible _get_product_price method within the Pricelist class

2. '**get_price' -> '_get_pack_line_price':** The get_price method on pack lines has been improved and renamed to  _get_pack_line_price. It is now capable of returning the price of the pack line for a specified pricelist.
    
3. **List Price Calculation:** Enhanced the _compute_product_lst_price method to return the list price of a pack. It now leverages a new method analogous to price_compute, named _pack_line_price_compute, which can determine the prices of pack components, including nested packs.

4. **Resolves an issue related to the action "Update Prices" on Sale Orders.** Now when update the pricelist just take into account the components of packs "Detailed - Detailed per component" and ignore the rest. More info about how to replicate the bug in the commit message.

5. **website Pack prices**: show the correct price on the website.
---


bruno-zanotti:
Hi @ernestotejeda, @pedrobaeza, @jjscarafia, @FrancoMaxime 👋🏼, what do you think about this PR? Although it is still a work in progress, the primary changes are already done. There is not huge modifications, just some minor refactor to fix some bugs and also to align our approach more closely on how Odoo handles product prices in lasts versions. I think this would be the base to get this modules working properly.. 
We're still testing (probably some adjustments will be needed, specially with different pricelists settings) and we have some remaining TODOs like improve the README and adapt website_sale_product_pack.
Any thoughts or feedback about this would be highly appreciated! 😃

pedrobaeza:
Pricelist support shouldn't be removed IMO, as we want specific prices for the components.

bruno-zanotti:
> Pricelist support shouldn't be removed IMO, as we want specific prices for the components.

Yes, I agree, but it seems weird to me to do it in price_compute.. right now without precilists is working fine, the next step will be test it with different pricelist settings and see how to adapt the code

pedrobaeza:
Well, the support should be in the base module, not in `sale_product_pack`, the same as in `product` against `sale`.

bruno-zanotti:
Hey @FrancoMaxime! Could you check this PR? Specifically focusing on the third commit related to the 'update prices' action. It is related to this commit https://github.com/OCA/product-pack/commit/68fce8c9d46eb88950a89d300916d16bf84629ef.

I'm no completely sure if it is a bug or if it was intentionally designed this way at the time of implementation, but I think that is a weird behavior. To clarify the issue, I'm attaching a couple of videos below. The first one is in the actual runboat and the second one with this PR applied.

In both cases I'm adding the same product before and after updating the priclist. However, in the first case the price of the products is different!

- Actual Runboat: [old.webm](https://github.com/OCA/product-pack/assets/58082586/22cc7df6-bed0-4274-adf1-69ce9537d635)
- After this PR: [new.webm](https://github.com/OCA/product-pack/assets/58082586/7347cddb-afe2-4ee6-90ef-8db5f5e5c600)

Thanks!

FrancoMaxime:
> Hey @FrancoMaxime! Could you check this PR? Specifically focusing on the third commit related to the 'update prices' action. It is related to this commit [68fce8c](https://github.com/OCA/product-pack/commit/68fce8c9d46eb88950a89d300916d16bf84629ef).
> 
> I'm no completely sure if it is a bug or if it was intentionally designed this way at the time of implementation, but I think that is a weird behavior. To clarify the issue, I'm attaching a couple of videos below. The first one is in the actual runboat and the second one with this PR applied.
> 
> In both cases I'm adding the same product before and after updating the priclist. However, in the first case the price of the products is different!
> 
>     * Actual Runboat: [old.webm](https://github.com/OCA/product-pack/assets/58082586/22cc7df6-bed0-4274-adf1-69ce9537d635)
> 
>     * After this PR: [new.webm](https://github.com/OCA/product-pack/assets/58082586/7347cddb-afe2-4ee6-90ef-8db5f5e5c600)
> 
> 
> Thanks!

Hello @Bruno-Zanotti, thank you for the care in your pull request. To be honest, if it something that i did, its a bug.
IMO, adding an item then apply a pricelist should give the same result as apply a pricelist then adding a product.
 I think you're doing great with your PR.  Its more readable, closer to the Odoo standard methods. 

for me, I validate these changes,  thanks a lot @Bruno-Zanotti

bruno-zanotti:
@FrancoMaxime thanks for checking and your comments!

bruno-zanotti:
Hey @ernestotejeda, @pedrobaeza, @jjscarafia, @FrancoMaxime, the refactor is finally finished, and IMO it's ready to merge. I've tried to make the commit messages and the PR description as clear as possible, detailing the changes and what I've tested. However, please let me know if you have any doubts or need further feedback.

product_pack/models/product_product.py
```diff
@@ -33,43 +33,16 @@ def split_pack_products(self):
     def price_compute(
         self, price_type, uom=False, currency=False, company=False, date=False
     ):
-        packs, no_packs = self.split_pack_products()
-        prices = super(ProductProduct, no_packs).price_compute(
+        packs, _ = self.split_pack_products()
+        prices = super(ProductProduct, self).price_compute(
             price_type, uom, currency, company, date
         )
+        # TODO:@bruno-zanotti prefetch_fields=False still necessary?
```
pedrobaeza: Finish this TODO

product_pack/tests/test_product_pack.py
```diff
@@ -75,7 +75,7 @@ def test_get_pack_lst_price(self):
         component_2.product_id.list_price = 15.0
         component_3 = self.env.ref("product_pack.pack_cpu_detailed_components_4")
         component_3.product_id.list_price = 5.0
-        self.assertEqual(50.0, pack.lst_price)
+        self.assertEqual(80.75, pack.lst_price)
```
pedrobaeza: Why this change?

product_pack/tests/test_product_pack.py
```diff
@@ -117,20 +117,3 @@ def test_pack_modifiable(self):
         pack.pack_type = "detailed"
         pack.pack_component_price = "totalized"
         self.assertTrue(pack.pack_modifiable_invisible)
-
-    def test_price_compute_with_pricelist_context(self):
```
pedrobaeza, line 121: Why this removal?

sale_product_pack/tests/test_sale_product_pack.py
```diff
@@ -98,7 +98,7 @@ def test_create_components_price_order_line(self):
         self.assertAlmostEqual(line.price_subtotal, 27.68)
         self.assertEqual(
             (self.sale_order.order_line - line).mapped("price_subtotal"),
-            [1755.0, 22.5, 885.0],
+            [1579.5, 20.25, 796.5],
```
pedrobaeza, line 101: Why this change?

pedrobaeza (changes requested):
The changes in the tests reveals that this is not backwards compatible and in fact is changing the way the prices are computed, so I think that's a stopper.

nicolascol:
Functional review is OK

product_pack/tests/test_product_pack.py
bruno-zanotti: This is because we are fixing the way the pack “totalized in main product” price is computed. According to the [README](https://github.com/OCA/product-pack/blob/3e8eb93b8431b5808aab2174b21ae357e70145f0/product_pack/README.rst?plain=1#L66-L68): the pack totalized price “will be its price plus the sum of all the components prices” but in the 16.0 the price is only the sum of the components. 
In the test:

- Sum of the components: 30+15+5 = 50.0
- Product Pack price: 30.75

Then the result of the test should be 50.0 + 30.75 = 80.75

product_pack/tests/test_product_pack.py
bruno-zanotti: That test was added in this commit d6044f1 basically to check if we are handling well the pricelist context in the [price_compute](https://github.com/OCA/product-pack/blob/3e8eb93b8431b5808aab2174b21ae357e70145f0/product_pack/models/product_product.py#L33) method. With this refactor we don’t need to handle the pricelist context since we are calling the odoo method “_get_contextual_price()” that does that for us (please check method _pack_line_price_compute()).

sale_product_pack/tests/test_sale_product_pack.py
bruno-zanotti: Before this PR the pricelist in the sale orders didn’t affect the pack components, which IMO is a bug. In the test, the pricelist “discount_pricelist” applies a 10% discount on all the products so we should modify the expected the components price. 
For example one of the components is "[FURN_9001] Flipover".

1. The product list price is: 1950. 
2. With the pricelist discount (10%): 1950 * 0.9 = 1755
3. With the PACK discount (10%): 1755 * 0.9 = 1579.50

Screenshot

bruno-zanotti:
> The changes in the tests reveals that this is not backwards compatible and in fact is changing the way the prices are computed, so I think that's a stopper.

Hey @pedrobaeza, thanks for the review! In my opinion, this is not only a refactor, we are also addressing some important bugs in the modules, as detailed in points 4 and 5 of the description and in the PR comments. That’s why I’m doing the PR to 16.0. However, I agree that this changes the module behavior so if you prefer we can move it to 17.0 (I already made the migration PRs so I can move it there).

pedrobaeza:
Well, in this case, the problem is not the version where this lands, but the change of concept. Doing in v15 only delays this change. When my customers arrive at v17, they will find that the prices have changed without any data change.

If you want to have the behavior of summing the prices of the components + the main product, create a new type of pack allowing to have that behavior, but without changing the current “Totalized in main product” one.

bruno-zanotti:
I disagree, IMO is not a change of concept, I think that this was always the idea of Totalized packs but at some point we introduce the bug (it is clear in the README). About the release:
1. When updating to a new mayor version a lot of things change and have to be adapted.
2. To keep the same behavior as before is as simple as putting the pack price in 0. We can even make a migration script to do that.

pedrobaeza:
> To keep the same behavior as before is as simple as putting the pack price in 0. We can even make a migration script to do that.

Can you absolutely confirm that in any case the price (pricelists included), the price of the parent product is going to be 0? I'm not sure.

bruno-zanotti:
> > To keep the same behavior as before is as simple as putting the pack price in 0. We can even make a migration script to do that.
> 
> Can you absolutely confirm that in any case the price (pricelists included), the price of the parent product is going to be 0? I'm not sure.

Yes it should work, this is why:

There are two possibilities:
- **Pricelist don’t affect the pack:** 
  - Before PR: price = components price
  - After PR: price = list pack price + components price
- **Pricelist affects the pack:** in this case it is important to understand that the pricelist applies on the SOL price (this is list pack price + components price)). 
  - Before PR: price = components * discount + fee
  - After PR: price = (list pack price + components) * discount + fee

Thus, normal totalized pack price **before** pr is the same as totalized pack price **after** pr when "list pack price" = 0

I've test it and this are the results (I did it in local database but you could check it on runboat): 
image

product_pack/models/product_pack_line.py
```diff
@@ -54,6 +54,20 @@ def _check_recursion(self):
...
-        return self.product_id.lst_price * self.quantity
+    def _pack_line_price_compute(
+        self, price_type, uom=False, currency=False, company=False, date=False
+    ):
+        pack_line_prices = {}
+        if self._context.get("pricelist"):
+            for line in self:
+                pack_line_prices[line.product_id.id] = line.with_context(
+                    quantity=line.quantity
+                ).product_id._get_contextual_price()
```
augusto-weiss:
```suggestion
                pack_line_prices[line.product_id.id] = line.product_id._get_contextual_price() * line.qunatity
```

augusto-weiss (changes requested):

FrancoMaxime:
Hi @bruno-zanotti,

I'm still reading every changes of this PR. Thank you again for all the work done.

bruno-zanotti:
Hey @FrancoMaxime !

Thanks for staying pending. It took us longer than expected, but I think we're finally nearing the end. We’ve adjusted our approach a bit (thanks to @augusto-weiss), removing the price_compute inheritance and instead handling price computation in _get_product_price, what I think it makes everything cleaner and more versatile code.

Also, we’ve separated issue https://github.com/OCA/product-pack/issues/169 into a new PR https://github.com/OCA/product-pack/pull/164 because it’s a stronger change so it needs to be handled properly. We’re planning to merge it into 17.0.

We are currently testing the PR on production databases, and if everything goes well, we should be ready to merge.

heroic:
this is currently an issue for us as well. product packs are only using the base price(not the tax) when pricing is not detailed per component, but if we use detailed per component then taxes are being included.

bruno-zanotti:
Hello @pedrobaeza @FrancoMaxime @nicolascol @heroic I close this PR since v16 is a relative stable version and move it to v17 here: https://github.com/OCA/product-pack/pull/173

PR #161, opened 2024-01-23
[16.0][FIX] product_pack: fix in price list search

lef-adhoc:
When searching for a price list through text, that is, without selecting a specific one in the search bar of the product.product model and not finding any, it triggers an initialization error for the variable pricelist when the "if pricelist..." condition is encountered.


petrus-v (approved):
code review ! LGTM

PR #173, opened 2024-06-27
[REF] x_product_pack: pricing refactor 

bruno-zanotti:
### Pricing refactor
This PR aims to perform a thorough refactor of the **product_pack, sale_product_pack and website_sale_product_pack** modules. The primary goal is to enhance the **pricing** logic for product packs, ensuring consistency and improved functionality.

Key improvements / fixes:

1. **Refactor Pack Price Computation:** Removed the inheritance from price_compute in favor of a clearer and
    more comprehensible _get_product_price method within the Pricelist class

2. '**get_price' -> '_get_pack_line_price':** The get_price method on pack lines has been improved and renamed to  _get_pack_line_price. It is now capable of returning the price of the pack line for a specified pricelist.
    
3. **List Price Calculation:** Enhanced the _compute_product_lst_price method to return the list price of a pack. It now leverages a new method analogous to price_compute, named _pack_line_price_compute, which can determine the prices of pack components, including nested packs.

4. **Resolves an issue related to the action "Update Prices" on Sale Orders.** Now when update the pricelist just take into account the components of packs "Detailed - Detailed per component" and ignore the rest. More info about how to replicate the bug in the commit message.

5. **website Pack prices**: show the correct price on the website.
---

PR #174, opened 2024-06-27
[17.0][MIG] website_sale_product_pack

augusto-weiss:


augusto-weiss:
@lef-adhoc @ioans73 @NICO-SOLUTIONS @pedrobaeza @bruno-zanotti 
Hello everyone!! Colud review this PR ??
Thanks all!!

bruno-zanotti (approved):
technical review LGTM

NICO-SOLUTIONS:
@augusto-weiss @bruno-zanotti  
looks quite good. except for one issue... Just a question of what should be the result.

When in define a Product-Pack with Pack Type "Detailed" & Component Price "Detailed per component".... Right now it displays in the webshop the price of the pack without the component prices on the product overiew. When you add this pack to the cart, it adds the component prices to the price of the pack. 
From my point of view it should be handled in the following way. When there is a product with Pack Type "Detailed" & Component Price "Detailed per component", it should show the sum of the comonents without adding the price of the pack to it in the product page. When i add the pack to the cart, the line with the pack should be zero (or maybe a note) and the prices of the components should be listed. 
So the behaviour would be closer to the "totalized in main product" scenario. 
The advantage of "Detailed per component" is, that you even can add products with different tax rates. This is not possible with "totalized in main product". 
So its about getting rid of the pack price itself for "Detailed per component" packs.

We should consider this here: #173

What do you think? If there is another purpose for this kind of pack, please let me know.

bruno-zanotti:
@NICO-SOLUTIONS 

Good point, that problem is also in older versions, so we decided to migrate the module as it is. Anyway, we made a PR in v17 with a pack's prices refactor that fixes lots of things included that issue, you can check it here https://github.com/OCA/product-pack/pull/173 (the fix is in the last commit)

NICO-SOLUTIONS (approved):
Technical review. LGTM

pedrobaeza:
/ocabot migration website_sale_product_pack

NOTE: On v18, probably product combos will be enough and this module family won't be needed.

NICO-SOLUTIONS:
> /ocabot migration website_sale_product_pack
> 
> NOTE: On v18, probably product combos will be enough and this module family won't be needed.

@pedrobaeza with combo choices you only can add each product with quantity = 1 to the choices. So i see the advantage of this module, where you can define the number of components. Havent check it in detail but for me this seem to be the bottelneck. Or am i missing something?

website_sale_product_pack/models/product_product.py
```diff
@@ -17,8 +17,9 @@ def check_website_published(self):
             if unpublished:
                 raise ValidationError(
                     _(
-                        "You can't add unpublished products (%(unpublished_products)s) to"
-                        "a published pack (%(pack_name)s)"
+                        "You can't add unpublished products "
+                        "(%(unpublished_products)s)"
+                        "to a published pack (%(pack_name)s)"
```
pilarvargas-tecnativa, line 22: Is this a pre-commit change?

pilarvargas-tecnativa (approved):

PR #179, opened 2024-09-04
[16.0][FIX] sale_product_pack: Define the correct sequence of sale order lines

victoralmau:
Define the correct sequence of sale order lines

Previously, the product pack line sequence was defined for the “extra” lines

@Tecnativa TT50677

PR #180, opened 2024-09-05
[17.0][FIX] sale_product_pack: Define the correct sequence of sale order lines

victoralmau:
FWP from 16.0: https://github.com/OCA/product-pack/pull/179

Define the correct sequence of sale order lines

Previously, the product pack line sequence was defined for the “extra” lines

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT50677

### DIFF
--- a/sale_product_pack/models/product_pack_line.py
+++ b/sale_product_pack/models/product_pack_line.py
@@ -16,7 +16,6 @@ class ProductPack(models.Model):
         quantity = self.quantity * line.product_uom_qty
         line_vals = {
             "order_id": order.id,
-            "sequence": line.sequence,
             "product_id": self.product_id.id or False,
             "pack_parent_line_id": line.id,
             "pack_depth": line.pack_depth + 1,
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

pedrobaeza (changes requested):
Uhm, wait, I'm thinking that this has a potential problem:

2 SO lines:

- Line 1 with pack. Sequence: 10
- Line 2. Sequence: 10

When unfolding line 1, the pack components will be behind line 2.

victoralmau:
The use case you indicate is already solved (actually for UI you cannot set several lines with the same sequence, it would be 10 and 11).

pedrobaeza:
On contrary, it's worst. The use case:

2 SO lines:

- Line 1 with pack with 2 sublines. Sequence: 10
- Line 2. Sequence: 11

When unfolding line 1, it will be:

- Line 1. Sequence: 10
- Subline 1. Sequence: 11
- Line 2. Sequence: 11
- Subline 2. Sequence: 12

victoralmau:
Sorry, I cannot reproduce the error you indicate.

ejemplo

victoralmau:
Best approach with https://github.com/OCA/product-pack/pull/187

PR #181, opened 2024-09-06
[16.0][FIX] stock_product_pack: Display Delivery Slip data correctly sorted in pack products

victoralmau:
Display Delivery Slip data correctly sorted in pack products

If any product pack has been used and is consumable we must reorder.
Example of picking:
- Consumable product (pack): sequence 10 (stock.move, id=1)
- Storable product A: sequence 11 (stock.move, id=2)
- Storable product B: sequence 12 (stock.move, id=3)

Confirm picking (Storable products have stock) and we get the following:
- stock.move.line, id=1, move_id=2 (Storable product A)
- stock.move.line, id=2, move_id=3 (Storable product B)
- stock.move.line, id=3, move_id=1 (Consumable product (pack))

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT50677

### DIFF
--- a/stock_product_pack/models/stock_move_line.py
+++ b/stock_product_pack/models/stock_move_line.py
@@ -1,0 +1,32 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    # TODO: Check if it is possible to remove it
+    def _get_aggregated_product_quantities(self, **kwargs):
+        """If any product pack has been used and is consumable we must reorder.
+        Example of picking:
+        - Consumable product (pack): sequence 10 (stock.move, id=1)
+        - Storable product A: sequence 11 (stock.move, id=2)
+        - Storable product B: sequence 12 (stock.move, id=3)
+        Confirm picking (Storable products have stock) and we get the following:
+        - stock.move.line, id=1, move_id=2 (Storable product A)
+        - stock.move.line, id=2, move_id=3 (Storable product B)
+        - stock.move.line, id=3, move_id=1 (Consumable product (pack))
+        Done picking and print Delivery Slip report, this method is used to print table,
+        and we get a confusing order, so we must reorder them according to the sequence
+        of the linked stock.move.
+        This is a known side effect in the stock addon when using consumable product +
+        storable products but for now we only reorder for product packs.
+        """
+        if any(
+            sml.product_id.pack_ok and sml.product_id.detailed_type == "consu"
+            for sml in self
+        ):
+            self = self.sorted(key=lambda x: x.move_id.sequence)
+        return super()._get_aggregated_product_quantities(**kwargs)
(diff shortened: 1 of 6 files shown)


sergio-teruel (approved):

stock_product_pack/models/stock_move_line.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    # TODO: Check if it is possible to remove it
```
pedrobaeza, line 10: This TODO is half explained. Remove it doing what? Link if there's Odoo PR, or mention that the problem is in Odoo, etc.

stock_product_pack/models/stock_move_line.py
```diff
@@ -0,0 +1,32 @@
...
+        and we get a confusing order, so we must reorder them according to the sequence
+        of the linked stock.move.
+        This is a known side effect in the stock addon when using consumable product +
+        storable products but for now we only reorder for product packs.
+        """
+        if any(
+            sml.product_id.pack_ok and sml.product_id.detailed_type == "consu"
+            for sml in self
+        ):
+            self = self.sorted(key=lambda x: x.move_id.sequence)
```
pedrobaeza, line 31: This is sorting everything, not only pack lines, so it can have side effects. I think this should only change the order of the pack lines to make sure they are next to the parent line. That can be done iterating through `self` and maybe adding lines to a new "ordered" recordset.

stock_product_pack/models/stock_move_line.py
victoralmau: I think the explanation inside the method is enough, do you need more? When I create a PR to solve this in odoo (it will take hours to review and try to solve the “real problem”) I will link it.

stock_product_pack/models/stock_move_line.py
victoralmau: I don't totally agree, I think trying to order “_just some lines_” can potentially be worse. 
- What if the sales order also had another consumable product (what order will it stay in)? 
- What if the sales order had some product removed from the pack? 
- What if the sales order had some product added to the pack in an exta way?

pedrobaeza: The question is that a specific module is modifying the overall delivery slip, which is not desirable.

For the cases you mention, 1 is totally out of the scope of this module, and 2 and 3 must be handled here or in `sale_stock_product_pack`.

victoralmau: In `stock_product_pack` it can't be done because we don't have the origin (`sale_line_id`), so ok, I will create a new module `sale_stock_product_pack` to solve these specific use cases.

All this does not happen in v14

pedrobaeza: In 13.0, this module exists: https://github.com/OCA/product-pack/tree/13.0/sale_stock_product_pack

victoralmau:
According to https://github.com/OCA/product-pack/pull/181#discussion_r1746809452 replaced with https://github.com/OCA/product-pack/pull/182

PR #184, opened 2024-09-16
[16.0][FIX] stock_product_pack: compute pack quantities

augusto-weiss:
Compute quantities for packs:

The error:
Could not finish a checkout process on website due to stock restriction
image

Reproduce error:
Create a pack.
Set pack as storable product, and set out of stock continue selling in FALSE
image
Make sure you have stock of pack lines.
Go to website and buy the pack
Try to finish the checkout process and you have the error

Also with this PR we ensure not to over-sell packs discriminating outgoing quantity in available quantity (line 37)

### DIFF
--- a/stock_product_pack/models/product_product.py
+++ b/stock_product_pack/models/product_product.py
@@ -13,31 +13,41 @@ class ProductProduct(models.Model):
     def _compute_quantities_dict(
         self, lot_id, owner_id, package_id, from_date=False, to_date=False
     ):
-        res = super()._compute_quantities_dict(
+        packs = self.filtered("pack_ok")
+        subproducts = packs.pack_line_ids.filtered(
+            lambda p: p.product_id.detailed_type == "product"
+        ).mapped("product_id")
+        res = super(ProductProduct, self | subproducts)._compute_quantities_dict(
             lot_id, owner_id, package_id, from_date=from_date, to_date=to_date
         )
-        packs = self.filtered("pack_ok")
-        for product in packs.with_context(prefetch_fields=False):
+        for pack in packs.with_context(prefetch_fields=False):
             pack_qty_available = []
             pack_virtual_available = []
             pack_free_qty = []
-            subproducts = product.pack_line_ids.filtered(
+
+            for line in pack.pack_line_ids.filtered(
                 lambda p: p.product_id.detailed_type == "product"
-            )
-            for subproduct in subproducts:
-                subproduct_stock = subproduct.product_id
-                sub_qty = subproduct.quantity
+            ):
+                sub_qty = line.quantity
                 if sub_qty:
                     pack_qty_available.append(
-                        math.floor(subproduct_stock.qty_available / sub_qty)
+                        math.floor(
+                            (
+                                res[line.product_id.id]["qty_available"]
+                                - res[line.product_id.id]["outgoing_qty"]
+                            )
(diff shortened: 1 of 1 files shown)


augusto-weiss:
Hello @pedrobaeza @victoralmau 
Could you check this issue, and let me know what do you think about the solution ?

pedrobaeza:
This must be handled in `website_sale_product_pack`, not here. You have also missed the pack setup, products configurations, etc.

augusto-weiss:
ok i understand, but I don't know if the problem has its origin on the website.

following the method _compute_quantities_dict the pack components return 0, and do not compute quantities correctly


About the pack I use non detailed (2nd img)

augusto-weiss:
Hi @pedrobaeza i moved  code to website_sale_product_pack.
Let me know your opinion ?  Thanks

pedrobaeza (commented):
Thanks for the changes, and sorry for not being very clear. I was expecting that you avoid the `_compute_quantities_dict` override, as that one is very general, but to focus on the website_sale checkout process itself, and bypass the stock checking just there on the met conditions. If the availability problem happens in other places, then it would be correct to put it on the main stock module, but then you should rephrase your rationale in the commit/PR message about how to reproduce it in a generic way (and maybe making reference that this happens also on e-commerce).

augusto-weiss:
Hi @pedrobaeza I was analyzing the issue in depth and following the call stack, I noticed that the compute_dict_qty method is called on the following line [(link),](https://github.com/odoo/odoo/blob/16.0/addons/website_sale_stock/models/sale_order.py#L65) so i should inherit the _get_cart_and_free_qty method.
But.. to do that I should add the dependency to website_sale_stock.

Let me know your opinion!

augusto-weiss:
`def _get_cart_and_free_qty(self, line=None, product=None, **kwargs):
        if (line and line.product_id.pack_ok) or (product and product.pack_ok):
            return 0, 0
        return super()._get_cart_and_free_qty(line=line, product=product, **kwargs)`

augusto-weiss:
```
def _get_cart_and_free_qty(self, line=None, product=None, **kwargs):
        if (line and line.product_id.pack_ok) or (product and product.pack_ok):
            return 0, 0
        return super()._get_cart_and_free_qty(line=line, product=product, **kwargs)
```

pedrobaeza:
Then it seems we need a `website_sale_stock_product_pack` glue module for keeping modularity. It's true that you can sell packs (of services, for example), without the need of having stocks. Anyway, why isn't this a problem on backend?

augusto-weiss:
Hi @pedrobaeza  Also happens in the backend... but by default is not blocking the possibility of confirming a sale order
image

Including the first approach (in stock_product_pack):
image

augusto-weiss:
Hello @pedrobaeza I take a step back to my first approach considering it must be computed correctly both in the backend also on the website

augusto-weiss:
@natuan9 @rven @rousseldenis  Hello everyone! Could someone review this pr?
Include two fixes:
1- Compute correctly pack quantities 
2- discriminate outgoing quantity when computing pack quantity available, to avoid oversell

Thanks for spending time!!

CC @pedrobaeza

pedrobaeza (approved):
Code review

/ocabot merge major

augusto-weiss:
@pedrobaeza Thanks for merge!!

PR #186, opened 2024-10-21
[16.0][FIX] sale_product_pack: Respect the order of the lines to be added

victoralmau:
Respect the order of the lines to be added

**Before**
antes

**After**
despues

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT51380

### DIFF
--- a/sale_product_pack/models/sale_order_line.py
+++ b/sale_product_pack/models/sale_order_line.py
@@ -75,18 +75,22 @@ class SaleOrderLine(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        new_vals = []
-        res = self.browse()
-        for elem in vals_list:
-            product = self.env["product.product"].browse(elem.get("product_id"))
-            if product and product.pack_ok and product.pack_type != "non_detailed":
+        """Only when strictly necessary (a product is a pack) will be created line
+        by line, this is necessary to maintain the correct order.
+        """
+        product_ids = [elem.get("product_id") for elem in vals_list]
+        products = self.env["product.product"].browse(product_ids)
+        if any(p.pack_ok and p.pack_type != "non_detailed" for p in products):
+            res = self.browse()
+            for elem in vals_list:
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
We are losing here a lot of performance, as this applies to all sales order lines, no matter if with pack or not.

Aren't there any other option for this?

victoralmau:
> We are losing here a lot of performance, as this applies to all sales order lines, no matter if with pack or not.
> 
> Aren't there any other option for this?

Changed so that it only applies when any line that is created is a pack.

pedrobaeza:
But it still has the problem with large orders containing packs. You have to find another alternative. How it was done in previous versions? What is the difference now?

victoralmau:
v15 code: https://github.com/OCA/product-pack/blob/15.0/sale_product_pack/models/sale_order_line.py#L79
After creating each of the lines, the extra lines of the pack were auto-created (if applicable).

victoralmau:
Ping @pedrobaeza

pedrobaeza:
Why don't you try to incorporate the pack lines to the dictionary vals list after each record, and call super with all of them, and then only returning those recordsets that are not pack lines? Do you understand me?

victoralmau:
> Why don't you try to incorporate the pack lines to the dictionary vals list after each record, and call super with all of them, and then only returning those recordsets that are not pack lines? Do you understand me?

Yes, I understand what you are saying but “_it is not possible_”. It would be to add to the dictionary the “_extra lines_” that are set in https://github.com/OCA/product-pack/blob/16.0/sale_product_pack/models/sale_order_line.py#L56 but those values are obtained precisely by taking as reference the “_parent_” line (`sale.order.line`) already existing https://github.com/OCA/product-pack/blob/16.0/sale_product_pack/models/product_pack_line.py#L14.

pedrobaeza (approved):
OK if there's no other solution then.

/ocabot merge patch

PR #209, opened 2025-05-12
[IMP] port https://github.com/OCA/product-pack/pull/159 to v17

damdam-s:
port https://github.com/OCA/product-pack/pull/159 to v17

PR #210, opened 2025-05-12
[17.0][MIG] sale_product_pack_contract

damdam-s:
depends on #209


alexandregaldeano (approved):
LGTM

sale_product_pack/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Sale Product Pack",
-    "version": "17.0.2.0.1",
+    "version": "17.0.2.0.2",
```
victoralmau, line 5: This change is not necessary if no migration scripts are added, when something is merged, _bot_ does the version change automatically.

PR #216, opened 2025-07-10
[18.0][MIG]purchase_product_pack: Migration to 18.0

matteotognini:


matteotognini:
@pedrobaeza Sorry to bother you.
The tests failed because the tracking of the product used in the test has changed here: https://github.com/odoo/odoo/blob/18.0/addons/purchase_stock/data/purchase_stock_demo.xml#L37
So a lot is required to confirm the picking.
How can I fix it? I tried to inherit the product data and change the tracking but it doesn't work.
Do I need to replace the product used in the tests?

pedrobaeza:
You should avoid to use demo data in your tests:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#avoid-relying-on-demo-data

That way, you define exactly the behavior of your records. You can switch them.

matteotognini:
Thanks! The test failure occurred in _sale_stock_product_pack_, not in this module. It only failed now because purchase_stock is a dependency of this module. Should I add a commit to update the tests in _sale_stock_product_pack_ avoiding using demo data?

pedrobaeza:
Yes, that would be great. I'm removing all demo data usage in tests on the modules I'm migrating exactly for this.

matteotognini:
Now the tests are green, I didn't replaced the demo data with test-created products, but I did check the tracking on the  pack components.I also used a different product pack to test the company switch, since the one previously used had already been moved and its company could no longer be changed.

pedrobaeza:
I insist in replacing demo data, as doing that is only a patch to a fragile situation.

matteotognini:
I know, but which ones exactly? Do I need to replace all the demo records that use "`product_pack_cpu_detailed_components`" as `parent_product_id`?
Or is it enough to just replace the component with one created during the test? Thanks!

pedrobaeza:
I would replace all the data used in the test, and reduce the test data to the minimum you need for testing the features. Why putting 5 pack lines while you may only need 1 or 2? Do you understand me?

pedrobaeza:
Hi @matteonext I have replaced demo data in #217 and #218 FYI. You can review it and rebase removing your changes when both are merged.

pedrobaeza:
You can rebase now and undo the changes in other modules tests.

matteotognini:
Hi @pedrobaeza, Thanks a lot for your help to avoid using demo data in tests. I removed the commits and rebased.

pedrobaeza (approved):
Let's merge it and I'll change tests to inherit from existing ones and without demo data in a subsequent PR:

/ocabot merge nobump

pedrobaeza:
Here you have it: #219

PR #217, opened 2025-07-15
[18.0][IMP] product_pack: Don't depend on demo data for tests + simplification

pedrobaeza:
Demo data can make tests to fail in integration environments or locally modified ones, so let's remove them and use data created on the test itself.

We also improve the containerization of the elements of the tests.

In the same movement, we have simplified the code of the test for doing the same with less lines and in a clearer way and change the old maintainer.

@Tecnativa


victoralmau (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #220, opened 2025-08-13
[18.0][FIX] product_pack: prevent currency being misinterpreted as UoM when printing labels for totalized packs

matiasperalta1:
Issue:
When printing product labels for pack products with Pack Components Price set to Totalized on the main product, Odoo raised:
AttributeError: 'res.currency' object has no attribute '_compute_quantity'
This happened because in some _get_product_price calls, the currency parameter was being passed as the second positional argument and misinterpreted as uom, causing the crash.

Solution:
Updated _get_product_price to detect when the second positional argument is a res.currency record, move it to kwargs['currency'], and set uom=None. This ensures _compute_price_rule receives the correct parameters and avoids using a currency as a UoM.

Step to reproduce error and test 
1. Create a pack product.
2. Set Pack Components Price to Totalized.
3. Add pack component lines.
4. From the product form, press Print Label. The error should not be appear

### DIFF
--- a/product_pack/models/product_pricelist.py
+++ b/product_pack/models/product_pricelist.py
@@ -25,7 +25,13 @@ class Pricelist(models.Model):
                 ][0]
 
             for line in product.sudo().pack_line_ids:
-                pack_price += line._get_pack_line_price(self, *args, **kwargs)
+                pack_price += line._get_pack_line_price(
+                    self,
+                    kwargs.get("quantity", args[0] if args else 1.0),
+                    uom=kwargs.get("uom"),
+                    date=kwargs.get("date"),
+                    currency=kwargs.get("currency"),
+                )
             return pack_price
         else:
             return super()._get_product_price(product, *args, **kwargs)


pedrobaeza:
Which crash? Please check pre-commit.

matiasperalta1:
@pedrobaeza I have already completed the description of the change

lav-adhoc (approved):
LGTM

lef-adhoc (approved):

pedrobaeza:
Isn't this a problem of the calling method not using keyword argument?

product_pack/models/product_pricelist.py
```diff
@@ -11,6 +11,7 @@ def _get_product_price(self, product, *args, **kwargs):
                   considering pricelist rules
         """
         self and self.ensure_one()
+
```
pedrobaeza: Don't add empty lines inside a method.

matiasperalta1:
@pedrobaeza Yes, the original problem was that the calling method didn't use keyword arguments for currency, causing the currency to be passed as a positional argument and ending up in the uom parameter.
I refactored the call to ensure all relevant arguments were passed as keyword arguments, thus avoiding the error and making the code clearer and safer.

pedrobaeza:
> Isn't this a problem of the calling method not using keyword argument?

What about my question?

matiasperalta1:
@pedrobaeza Yes, exactly. The root cause was the calling method not using keyword arguments. The fix I pushed addresses that by refactoring the call to use explicit keywords.

pedrobaeza:
Then this is not correct. You shouldn't fix things at other level that doesn't correspond, as this is complicating this code in an unneeded way. Please fix it where the keyword argument is not being called properly (even at Odoo level).

Imagine if other keyword argument is also not passed as keyword (different from `currency_id`), or the order is not exactly the same, you will add more code for handling that? I think that's not the proper fix.

I understand it's easier to do it at this level, but think that all the modules will expect the same things in `kwargs`, and this patch may come more times in other modules, so fix it at the root level.

pedrobaeza (changes requested):
To be fixed in the root.

matiasperalta1:
@pedrobaeza Thanks for the feedback!
I really appreciate your guidance about not patching the issue at the wrong level. I investigated further and addressed the root cause directly in Odoo in this PR: https://github.com/odoo/odoo/pull/223315

pedrobaeza:
Fixed in Odoo, so closing.

PR #221, opened 2025-09-25
support for Odoo 19

hyperjump888:
The "is a pack" option did not show up in Odoo 19 due to UI changes. It has been fixed to show up now in Odoo 19.


pedrobaeza:
When the 19.0 branch is created (it will be soon), please do the complete migration following the migration guide.

PR #233, opened 2025-11-12
[16.0][FIX] sale_stock_product_pack: compute qty_to_invoice for pack parents on returns

lef-adhoc:
When pack_component_price is "totalized" or "ignored", child lines have price_unit=0 (price is on parent). On product returns, only children update qty_to_invoice, parent stays at 0.
Credit notes only include lines with qty_to_invoice!=0, so only children appear with price=0, resulting in CN amount 0.

Fix: calculate parent's qty_to_invoice based on children's minimum, same logic as qty_delivered.

### DIFF
--- a/sale_stock_product_pack/models/sale_order.py
+++ b/sale_stock_product_pack/models/sale_order.py
@@ -29,3 +29,23 @@ class SaleOrderLine(models.Model):
                 delivered_packs.append(pack_line.qty_delivered / qty_per_pack)
             line.qty_delivered = delivered_packs and min(delivered_packs) or 0.0
         return res
+
+    def _compute_qty_to_invoice(self):
+        res = super()._compute_qty_to_invoice()
+        pack_parent_lines = self.filtered(
+            lambda x: x.pack_child_line_ids
+            and x.product_id.pack_ok
+            and x.state in ["sale", "done"]
+        )
+        for line in pack_parent_lines:
+            child_qty_to_invoice = []
+            for pack_line in line.pack_child_line_ids.filtered("product_uom_qty"):
+                if line.product_uom_qty:
+                    qty_per_pack = pack_line.product_uom_qty / line.product_uom_qty
+                    if qty_per_pack:
+                        packs_to_invoice = pack_line.qty_to_invoice / qty_per_pack
+                        child_qty_to_invoice.append(packs_to_invoice)
+
+            if child_qty_to_invoice:
+                line.qty_to_invoice = min(child_qty_to_invoice)
+        return res


lef-adhoc:
@pedrobaeza Hi, what do you think about this PR?

nicolascol (approved):
LGTM functional review
@pedrobaeza Could you merge please?

https://drive.google.com/file/d/1QIaj3b9m7P7pDftBEa1i4_Pd-yTIPyYc/view

sale_stock_product_pack/models/sale_order.py
```diff
@@ -29,3 +29,26 @@ def _compute_qty_delivered(self):
                 delivered_packs.append(pack_line.qty_delivered / qty_per_pack)
             line.qty_delivered = delivered_packs and min(delivered_packs) or 0.0
         return res
+
+    def _compute_qty_to_invoice(self):
+        res = super()._compute_qty_to_invoice()
+
```
pedrobaeza: Don't put empty lines inside a method.

pedrobaeza (commented):
Anyway, I don't understand the problem. The invoiceability is not by price, but on quantity.

lef-adhoc:
> Anyway, I don't understand the problem. The invoiceability is not by price, but on quantity.

@pedrobaeza  Here is a video replicating the problem: https://drive.google.com/file/d/1YIEe2FsTOuM0o9RFbnnRNc6u3wwzwG1j/view?usp=sharing

sale_stock_product_pack/models/sale_order.py
lef-adhoc: Deleted!

sale_stock_product_pack/models/sale_order.py
```diff
@@ -29,3 +29,23 @@ def _compute_qty_delivered(self):
...
+        )
+        for line in pack_parent_lines:
+            child_qty_to_invoice = []
+            for pack_line in line.pack_child_line_ids.filtered("product_uom_qty"):
+                if line.product_uom_qty:
+                    qty_per_pack = pack_line.product_uom_qty / line.product_uom_qty
+                    if qty_per_pack:
+                        packs_to_invoice = pack_line.qty_to_invoice / qty_per_pack
+                        child_qty_to_invoice.append(packs_to_invoice)
+
```
pedrobaeza, line 48: There's also this empty line. I spoke in general.

pedrobaeza (commented):
Thanks for the details, Franco. I think the weird thing is about that "Don't create move" check. Isn't this provoking to not create a line in the picking? Originally, the packs didn't have that option, and I consider it an alien option. This was originally intended to have the pack as consumable, so this problem didn't happen. Now I see that accepting the check was a mistake, as it provokes all these side effects. I think you should extract this logic into a separate module, and then do all the needed stuff there, but without contaminating the usual flows.

nicolascol:
Hi Pedro. In v18 I’m not able to reproduce this error. I also wanted to mention that this configuration of "Don't create move" is useful because it allows selecting storable packs and therefore displaying stock in the ecommerce.

pedrobaeza:
Well, I think that can be done with consumable as well, but let's not block it.

/ocabot merge patch

PR #241, opened 2026-01-27
[18.0][IMP] sale_product_pack: improve ergonomy for the user experience

StephaneMangin:
Ergonomy is a bit confusing for the customer as constrains on changes and especially unlink are based only on onchange mechanism (which doesn't properly update the view to reflect the final result)

So the solution to allow a better readability of what the customer is doing is to:
 - Convert the component lines unlink exception in JS
 - Also remove component lines in JS when a product pack is removed

Also add some new tests for component lines forbidden changes


sale_product_pack/models/sale_order_line.py
```diff
@@ -100,15 +110,27 @@ def write(self, vals):
...
+                x.pack_parent_line_id not in self
+                or x.pack_parent_line_id in x.order_id.order_line
+            )
+            and not x.pack_parent_line_id.product_id.pack_modifiable
+        ):
+            raise UserError(
+                _(
+                    "You cannot delete this line because is part of a pack in"
+                    " this sale order. In order to delete this line you need to"
+                    " delete the pack itself"
```
yvaucher:
```suggestion
                    "You cannot delete this line because it is part of a pack in"
                    " this sale order. In order to delete this line you need to"
                    " delete the pack itself."
```

Small syntax issue

yvaucher (commented):
Code LGTM, only the error message that has an error.

StephaneMangin:
> Tests are failing, this might require to make a check bypass with test on `config["test_enable"]`

Yes, i'm on it. Sale order copy must be adapted (removing children lines from the copy data for instance)

yvaucher:
Please check the unit tests which are failing on:

TestSaleOrder.test_unlink_multiple_packs and TestSaleOrder.test_unlink_pack_line

Where the exception is raised when it's not expected.

StephaneMangin:
> Please check the unit tests which are failing on:
> 
> TestSaleOrder.test_unlink_multiple_packs and TestSaleOrder.test_unlink_pack_line
> 
> Where the exception is raised when it's not expected.

All good now. Refactoring the unlink process and the order liens copy (adding a context option to bypass children unlink contrains)

yvaucher (approved):

PR #225, opened 2025-10-20
[19.0] [MIG] sale_product_pack : migrate to 19.0

apik-mgu:

PR #244, opened 2026-03-12
[19.0] [MIG] sale_product_pack : migrate to 19.0

FrancoMaxime:
Its the same migration as #225
But taking into account the remark of @pedrobaeza and @ivantodorovich


sale_product_pack/models/product_pack_line.py
```diff
@@ -25,7 +25,6 @@ def get_sale_order_line_vals(self, line, order):
             "product_uom_qty": quantity,
         }
         sol = line.new(line_vals)
```
pedrobaeza: As now all the fields are computed writable, this technique is not needed anymore, and lines can be created with the minimum data in the dictionary, so please switch to this new approach.

FrancoMaxime: I added an overload of the _compute_name method to add the product_pack name's behavior

I fixed too a way to bypass the onchange constraint

nicolascol (approved):
LGTM
Could you merge please?
@pedrobaeza

sale_product_pack/models/product_pack_line.py
pedrobaeza: This is not what I was talking about. I was talking about creating the record directly, without `new`, onchanges, and so on.

pedrobaeza (changes requested):

nicolascol:
Hi @FrancoMaxime 
Did you see Pedro's comment?

FrancoMaxime:
I don't know why i can't answer directly in the @pedrobaeza comment's.
The new method is no longer used, i just added a fix for a strange behavior when you're playing with multiple tabs of your browser.

As we decided to remove the description creation in the method get_sale_order_line_vals, i had to overwrite the compute_name function instead to keep the behavior.

If i'm missing something, could you please give me more information about what to do.

Thanks you @pedrobaeza

lef-adhoc:
@pedrobaeza Could you take a look at @FrancoMaxime 's comment?

pedrobaeza (approved):
/ocabot merge nobump

PR #246, opened 2026-04-08
[19.0][FIX] product_pack: fix wrong digits reference on pack line quantity

### LINKED ISSUE #245: Wrong Product Pack digits reference
The digits reference does not exist anymore. It is not possible to change it to more digits because of this.

## Module
product_pack

## Describe the bug
The digits reference does not exist anymore. It is not possible to change it to more digits because of this.

## To Reproduce
**Affected versions**:
19.0

**Expected behavior**
If UOM digits gets increased, pack line quantity should have more digits too.

**Additional context**
https://github.com/OCA/product-pack/blob/344a16a66be084bb627802f14e61cbaa16ee7d9b/product_pack/models/product_pack_line.py#L23 #

jappi00:
## Summary

- Replace non-existent `"Product UoS"` digits reference with `"Product Unit of Measure"` on `product.pack.line` quantity field
- Bump version to `19.0.1.0.1`
- Add test to verify the digits reference is valid

Closes #245

### DIFF
--- a/product_pack/models/product_pack_line.py
+++ b/product_pack/models/product_pack_line.py
@@ -20,7 +20,7 @@ class ProductPackLine(models.Model):
     quantity = fields.Float(
         required=True,
         default=1.0,
-        digits="Product UoS",
+        digits="Product Unit",
     )
     product_id = fields.Many2one(
         "product.product",


jappi00:
The topic was removed during the transition from Odoo 8.0 to Odoo 9.0: https://github.com/odoo/odoo/commit/d152106b1b8ee8294000bbca7d2ccc9c80a8d05e

jappi00:
@pedrobaeza is the test fine? I'am not sure but i want to make sure this not happen again. Is a very old bug.

jappi00:
I have created prs for some other versions. Not sure if it is the right way to do it? If so i would create the whole version PRs :D

product_pack/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Product Pack",
-    "version": "19.0.1.0.0",
```
pedrobaeza, line 5: Don't change version number. It will be done by the bot on merge. This way, you avoid unneeded conflicts.

product_pack/tests/test_product_pack.py
```diff
@@ -89,6 +89,13 @@ def test_pack_price_with_pricelist_context_ignored(self):
         )._get_contextual_price()
         self.assertEqual(price, 9)  # 10 with 10% discount
 
+    def test_pack_line_quantity_digits(self):
+        """Check that pack line quantity uses 'Product Unit' precision."""
+        dp = self.env["decimal.precision"].precision_get("Product Unit")
```
pedrobaeza: This test isn't effective, as if tomorrow, the precision unit disappears, it will be green anyway. On this case, better to not include a test.

jappi00: Done :D

pedrobaeza (approved):

tom-kramer (commented):
lgtm

tom-kramer (approved):

PR #254, opened 2026-04-29
[19.0][MIG] product_pack_type: Migration to 19.0

RaphaelVicini:
Based on the pull request of @rousseldenis: https://github.com/OCA/product-pack/pull/62
and on this PR of product_pack module: https://github.com/OCA/product-pack/pull/255


product_pack_type/models/product_pack_type.py
```diff
@@ -0,0 +1,41 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductPackType(models.Model):
+    _name = "product.pack.type"
+    _description = "Product Pack Type"
+
+    name = fields.Char()
```
FrancoMaxime: Maybe we should add a constraint to avoid duplicate

FrancoMaxime (approved):

BhaveshHeliconia (changes requested):
@RaphaelVicini The commit history doesn't look correct. There are commits related to other modules, but this PR should only contain commits for `product_pack_type`.

Could you please clean up the commit history so that it only includes the relevant commits for this module?

BhaveshHeliconia (commented):
There are still commits related to the product_pack module in the history. This PR should only contain commits for product_pack_type.

RaphaelVicini:
Hi, I need those commits because they include the changes that turned Product Pack into an app.

PR #255, opened 2026-04-30
[19.0][IMP] product_pack: Change module to application

RaphaelVicini:
Based on the PR of @rousseldenis (https://github.com/OCA/product-pack/pull/60)


rousseldenis (approved):

product_pack/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Product Pack",
-    "version": "19.0.1.0.2",
+    "version": "19.0.1.0.0",
```
FrancoMaxime: I think we should never downgrade the version outside migration.

What do you think @rousseldenis ?

FrancoMaxime (changes requested):

pedrobaeza (changes requested):
I don't think this should have first level menu, as this is only master data, not a traversal application which its complete flows.

PR #253, opened 2026-04-29
[16.0][FIX] sale_stock_product_pack: removed _get_aggregated_product_quantities

eduezerouali-tecnativa:
cc @Tecnativa TT62189

Before this change 
imagen
imagen

After the change
imagen
ping @pedrobaeza @victoralmau

PR #182, opened 2024-09-06
[16.0][MIG] sale_stock_product_pack: Migration to 16.0

victoralmau:
Changes done:
- [x] Migration to 16.0
- [x] Display Delivery Slip data correctly sorted in pack products (https://github.com/OCA/product-pack/pull/181)

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT50677

PR #170, opened 2024-05-17
[MIG] sale_stock_product_pack: Migration to 17.0

bruno-zanotti:

PR #258, opened 2026-06-11
[16.0][IMP+FIX] sale_stock_product_pack + stock_product_pack: Display Delivery Slip data correctly sorted in pack products

victoralmau:
Display Delivery Slip data correctly sorted in pack products

Related to https://github.com/OCA/product-pack/pull/182/changes/63e445ac02276bcbd2cd44a74e1457b5cb4c05cf + https://github.com/OCA/product-pack/pull/253

Changes done:
- [x] The `_get_aggregated_product_quantities()` method has been simplified; if there are any "pack moves", the `stock.move.line` records are sorted according to the `stock.move` order so that the data appears in the correct order
- [x] The `_get_aggregated_product_quantities()` method has been moved to `stock_product_pack` so that it works in all cases (sales and purchases)
- [x] The `_get_aggregated_properties()` method is overridden to specify a different `line_key`, ensuring that data is not incorrectly grouped if there are extra lines using any of the products contained in the pack
- [x] Tests have been added to cover more complex use cases (component1 + component2 + product_pack) data

This same approach (the `_get_aggregated_properties()` method) could be used in a `purchase_stock_product_pack` module

Please @pedrobaeza and @eduezerouali-tecnativa can you review it?

@Tecnativa TT62930 TT62189

### DIFF
--- a/sale_stock_product_pack/models/stock_move_line.py
+++ b/sale_stock_product_pack/models/stock_move_line.py
@@ -1,4 +1,4 @@
-# Copyright 2024 Tecnativa - Víctor Martínez
+# Copyright 2024-2026 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import models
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Please do it directly in 18.

victoralmau:
> Please do it directly in 18.

This isn't possible because the change added in https://github.com/OCA/product-pack/pull/182 wasn't included in 17 https://github.com/OCA/product-pack/pull/170 or 18; therefore, in my opinion, the first step is to fix this behavior in 16 and then add it to later versions.

pedrobaeza:
I don't see the problem. Just add everything needed in 18, being one commit or several. We don't need it either for 16 nor 18, and now you have to deal with other things in these old versions, like pre-commit.

victoralmau:
Moved to https://github.com/OCA/product-pack/pull/259

PR #262, opened 2026-07-03
[FIX] stock_product_pack: don't skip non detailed packs on dont_create_move

gal-adhoc:
## Summary

`stock.rule.run()` only removed the parent pack's procurement (so it doesn't create its own `stock.move`) when `pack_type == "detailed"`. A **non detailed** pack with `dont_create_move` active kept generating a `stock.move` for the parent product in the delivery, defeating the purpose of the flag.

Additionally, `_compute_qty_delivered` derives a pack line's own `qty_delivered` from its components, but only depended on the pack line's **own** (untouched) stock moves — so returning a pack's components never invalidated the cached value on the parent, leaving it stale.

## Change

- **`stock_product_pack/models/stock_rule.py`**: the exclusion now depends only on `pack_ok` + `dont_create_move`, applying to both `detailed` and `non_detailed` packs. Also avoids mutating the `procurements` list while iterating over it (the previous `for ... procurements.remove(...)` pattern could skip elements).
- **`sale_stock_product_pack/models/sale_order.py`**: added the same `@api.depends` `sale_stock` declares for `qty_delivered`, one hop further through the components, so a pack's delivered quantity is correctly recomputed when its components are returned.

`qty_delivered` is deliberately left driven by real stock moves only. A non detailed pack that moves nothing stays at `Delivered = 0`: such a pack shouldn't be invoiced on delivery in the first place — that is what `invoice_policy` is for. (This contradicts `dont_create_move`'s current help text, which promises the pack is "set as delivered upon sale confirmation"; see the discussion below.)

## Test plan

- `test_dont_create_move_non_detailed_pack`: confirms an order with a non detailed pack (`dont_create_move=True`), asserts the delivery has no `stock.move` for the pack product and that `qty_delivered` stays at 0.
- `test_enabling_dont_create_move_keeps_delivery_history`: sells 4 packs with the flag off, delivers 2, then enables the flag — asserts the delivered quantity is not restated.
- `test_qty_delivered_pack_after_component_return`: sells 2 detailed packs, delivers, returns exactly 1 pack's worth of both components proportionally, asserts `qty_delivered` drops accordingly without any manual recompute.
- All three verified to fail against the code before their respective fix.
- Full `product_pack` / `sale_product_pack` / `stock_product_pack` / `sale_stock_product_pack` suites pass: **29 tests, 0 failures**.

### DIFF
--- a/sale_stock_product_pack/models/sale_order.py
+++ b/sale_stock_product_pack/models/sale_order.py
@@ -1,12 +1,25 @@
 # Copyright 2021 Tecnativa - David Vidal
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 # pylint: disable=W8110
-from odoo import models
+from odoo import api, models
 
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
+    # Same depends sale_stock declares for qty_delivered [1], just one hop
+    # further through the components: a pack line's own qty_delivered is
+    # derived from its children's qty_delivered below, so it must be
+    # invalidated/recomputed whenever a child's own moves change (e.g. a
+    # return), not only when the pack line's own moves do.
+    # [1] https://github.com/odoo/odoo/blob/19.0/addons/sale_stock/models/
+    #     sale_order_line.py -> SaleOrderLine._compute_qty_delivered
+    @api.depends(
+        "pack_child_line_ids.move_ids.state",
+        "pack_child_line_ids.move_ids.location_dest_usage",
+        "pack_child_line_ids.move_ids.quantity",
+        "pack_child_line_ids.move_ids.product_uom",
+    )
     def _compute_qty_delivered(self):
         """Compute pack delivered pack quantites according to its components
         deliveries"""
--- a/sale_stock_product_pack/tests/test_sale_stock_product_pack.py
+++ b/sale_stock_product_pack/tests/test_sale_stock_product_pack.py
@@ -136,3 +136,162 @@ class TestSaleStockProductPack(TestSaleProductPackBase):
         )
         self.assertEqual(line_pack_component2_data["qty_ordered"], 2)
         self.assertEqual(line_pack_component2_data["quantity"], 2)
+
+    def test_dont_create_move_non_detailed_pack(self):
+        """A 'Non Detailed' pack with 'dont_create_move' active must not
+        generate a stock.move for the parent product either.
+        Regression test for stock_rule.run(), which used to only skip the
+        parent procurement when pack_type == 'detailed', silently ignoring
+        the flag for non detailed packs.
+
+        qty_delivered is left untouched: it stays driven by real stock
+        moves, so a pack that moves nothing is simply not delivered. Such a
+        pack shouldn't be invoiced on delivery in the first place -- that
+        is what invoice_policy is for.
+        """
+        self.pack.write(
+            {
+                "pack_type": "non_detailed",
+                "dont_create_move": True,
+            }
+        )
+        pack_line = self._add_so_line()
+        self.sale = self.sale_order
+        self.sale.action_confirm()
+        self.assertFalse(
+            self.sale.picking_ids.move_ids.filtered(
+                lambda move: move.product_id == self.pack
+            ),
+            "The pack product must not generate a stock.move when "
+            "'dont_create_move' is active, regardless of its pack display "
+            "type.",
+        )
+        self.assertEqual(
+            pack_line.qty_delivered,
+            0.0,
+            "Confirming the order must not mark the pack as delivered: "
+            "qty_delivered has to reflect real stock moves only.",
+        )
+
+    def test_enabling_dont_create_move_keeps_delivery_history(self):
+        """Enabling 'dont_create_move' on a product must not rewrite the
+        delivered quantity of orders that were already (partially)
+        delivered.
+
+        Regression test for the scenario reported in review: sell 4 packs
+        with the flag disabled, deliver 2, then enable the flag ->
+        qty_delivered used to jump from 2 to 4, altering delivery history
+        and risking over-invoicing.
+        """
+        self.pack.pack_type = "non_detailed"
+        pack_line = self._add_so_line()
+        pack_line.product_uom_qty = 4
+        self.sale_order.action_confirm()
+        # Deliver 2 out of the 4 ordered packs.
+        pack_move = self.sale_order.picking_ids.move_ids.filtered(
+            lambda move: move.product_id == self.pack
+        )
+        self.assertTrue(pack_move, "The pack should move while the flag is off.")
+        pack_move.quantity = 2
+        pack_move.picked = True
+        pack_move.picking_id._action_done()
+        self.assertEqual(pack_line.qty_delivered, 2.0)
+
+        self.pack.dont_create_move = True
+
+        self.assertEqual(
+            pack_line.qty_delivered,
+            2.0,
+            "Enabling 'dont_create_move' must not restate what was already delivered.",
+        )
+
+    def _create_return_moves(self, delivery_moves, qty_by_product):
+        # Return moves are created directly via the ORM (state='done',
+        # to_refund=True) instead of going through stock.picking /
+        # stock.return.picking: that orchestration depends on the
+        # warehouse's delivery route (one-step vs multi-step) and on
+        # enterprise overrides (e.g. helpdesk_stock redeclares the wizard's
+        # 'picking_id' as a computed field with unrelated resolution
+        # logic), both of which make it fragile and environment-dependent
+        # for a unit test. This mirrors exactly the fields
+        # _compute_qty_delivered's filter looks at.
+        return_picking = self.env["stock.picking"].create(
+            {
+                "picking_type_id": delivery_moves[:1].picking_id.picking_type_id.id,
+                "location_id": delivery_moves[:1].location_dest_id.id,
+                "location_dest_id": delivery_moves[:1].location_id.id,
+            }
+        )
+        for move in delivery_moves:
+            qty = qty_by_product.get(move.product_id, 0.0)
+            if not qty:
+                continue
+            self.env["stock.move"].create(
+                {
+                    "product_id": move.product_id.id,
+                    "product_uom_qty": qty,
+                    "quantity": qty,
+                    "product_uom": move.product_uom.id,
+                    "picking_id": return_picking.id,
+                    "location_id": move.location_dest_id.id,
+                    "location_dest_id": move.location_id.id,
+                    "sale_line_id": move.sale_line_id.id,
+                    "origin_returned_move_id": move.id,
+                    "to_refund": True,
+                    "state": "done",
+                    "picked": True,
+                }
+            )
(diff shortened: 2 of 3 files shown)


sale_stock_product_pack/models/sale_order.py
```diff
@@ -1,12 +1,32 @@
...
+    # qty_delivered below, so it must be invalidated/recomputed whenever a
+    # child's own moves change (e.g. a return), not just when the pack
+    # line's own moves do.
+    # The last 4 depends mirror sale_stock's own depends on qty_delivered
+    # (move_ids.*), just one hop further through the components: a detailed
+    # pack line's own qty_delivered is derived from its children's
+    # qty_delivered below, so it must be invalidated/recomputed whenever a
+    # child's own moves change (e.g. a return), not just when the pack
+    # line's own moves do.
+    @api.depends(
```
Sadiq-OSI, line 17: Please add `product_uom_qty` to this `@api.depends` list and add a regression test for editing the quantity after confirmation.

`qty_delivered` is assigned from `product_uom_qty` below, but changing a confirmed line from 2 units to 5 currently leaves `qty_delivered` at 2 because the stored compute is not invalidated. With invoicing based on delivered quantities, this can produce an incorrect invoice quantity.

sale_stock_product_pack/models/sale_order.py
```diff
@@ -29,3 +49,18 @@ def _compute_qty_delivered(self):
...
+        # A pack with 'dont_create_move' active never gets its own stock.move
+        # (see stock_product_pack's stock.rule override), so it can't be
+        # marked delivered through the regular stock_move-based computation.
+        # This only applies to packs without pack_child_line_ids (e.g. non
+        # detailed packs): detailed packs are already handled above, deriving
+        # their delivered quantity from their components. Consider the
+        # former delivered upon confirmation, as dont_create_move's field
+        # help text documents.
+        for line in self.filtered(
+            lambda x: x.qty_delivered_method == "stock_move"
```
Sadiq-OSI: Please restrict this assignment to confirmed sale lines, for example by requiring `x.state == "sale"`, and add a draft-order regression test.

The current condition applies in every state. If `dont_create_move` is enabled for a pack already used on a draft quotation, the dependency recompute changes `qty_delivered` from 0 to the full ordered quantity before confirmation. That contradicts the field help, which says the pack is delivered upon sale confirmation.

Sadiq-OSI (changes requested):
The procurement filtering itself looks correct, and the focused module tests pass.

I found two delivered-quantity edge cases that should be fixed before merging:

- Changing the ordered quantity after confirmation does not recompute `qty_delivered`.
- A draft quotation can be marked delivered when `dont_create_move` is enabled.

Please add regression coverage for both scenarios.

sale_stock_product_pack/models/sale_order.py
gal-adhoc: Fixed — added `product_uom_qty` to the `@api.depends` list. Added a regression test (`test_dont_create_move_qty_change_after_confirm`) that confirms a line, then bumps `product_uom_qty` from 1 to 5, and asserts `qty_delivered` follows it. Verified it fails without the depends fix.

sale_stock_product_pack/models/sale_order.py
gal-adhoc: Fixed — added `x.state == "sale"` to the filter. Added a regression test (`test_dont_create_move_draft_quotation_not_delivered`) that adds the line while draft, then toggles `dont_create_move` on (triggering the dependency recompute), and asserts `qty_delivered` stays 0 in both cases. Verified it fails without the state check.

sale_stock_product_pack/models/sale_order.py
```diff
@@ -1,12 +1,28 @@
...
+    # line's own moves do. product_uom_qty is needed because the
+    # dont_create_move branch assigns qty_delivered from it directly.
+    @api.depends(
+        "state",
+        "product_uom_qty",
+        "product_id.dont_create_move",
+        "pack_child_line_ids.move_ids.state",
+        "pack_child_line_ids.move_ids.location_dest_usage",
+        "pack_child_line_ids.move_ids.quantity",
+        "pack_child_line_ids.move_ids.product_uom",
```
victoralmau:
```suggestion
       # comment: 
        "pack_child_line_ids.move_ids.state",
        "pack_child_line_ids.move_ids.location_dest_usage",
        "pack_child_line_ids.move_ids.quantity",
        "pack_child_line_ids.move_ids.product_uom",
```
I think it would be interesting to point out that these are the same `depends` used by `sale_stock`.: https://github.com/odoo/odoo/blob/dacaad91bba8f959daf5d89a046c5a1c11e48eec/addons/sale_stock/models/sale_order_line.py#L195

sale_stock_product_pack/models/sale_order.py
```diff
@@ -1,12 +1,28 @@
...
+    # (move_ids.*), just one hop further through the components: a detailed
+    # pack line's own qty_delivered is derived from its children's
+    # qty_delivered below, so it must be invalidated/recomputed whenever a
+    # child's own moves change (e.g. a return), not just when the pack
+    # line's own moves do. product_uom_qty is needed because the
+    # dont_create_move branch assigns qty_delivered from it directly.
+    @api.depends(
+        "state",
+        "product_uom_qty",
+        "product_id.dont_create_move",
```
victoralmau: IMO, `product_id.dont_create_move` shouldn't be in the `depends` list; if the value is changed for orders that have already been fully done, all order lines for that product shouldn't be recalculated.

sale_stock_product_pack/models/sale_order.py
```diff
@@ -1,12 +1,28 @@
...
+    # The last 4 depends mirror sale_stock's own depends on qty_delivered
+    # (move_ids.*), just one hop further through the components: a detailed
+    # pack line's own qty_delivered is derived from its children's
+    # qty_delivered below, so it must be invalidated/recomputed whenever a
+    # child's own moves change (e.g. a return), not just when the pack
+    # line's own moves do. product_uom_qty is needed because the
+    # dont_create_move branch assigns qty_delivered from it directly.
+    @api.depends(
+        "state",
+        "product_uom_qty",
```
victoralmau:
```suggestion
```
IMO, `product_uom_qty` shouldn't be included in the `depends` list. If that value is changed, the automatic updates made in `stock.move` will ensure that the `depends` for `pack_child_line_ids` are sufficient.

victoralmau (changes requested):

sale_stock_product_pack/models/sale_order.py
dcanete-pyming: Confirmed functionally on the PR Runboat: I sold 4 non-detailed packs with dont_create_move disabled and delivered 2. After enabling the flag on the product, the sale order's delivered quantity changed retroactively from 2 to 4. This modifies the delivery history and may result in over-invoicing.

dcanete-pyming (changes requested):

sale_stock_product_pack/models/sale_order.py
gal-adhoc: Thanks for catching this — you were right, and it was the clearest symptom of a deeper design problem.

The root cause was a block I had added that wrote `qty_delivered = product_uom_qty` on confirmation for packs with `dont_create_move`. Writing a stock-driven field from order state is what allowed the retroactive restatement you reproduced. **I removed that block entirely**, so `qty_delivered` is now driven by real stock moves only.

Added `test_enabling_dont_create_move_keeps_delivery_history` covering exactly your scenario (sell 4, deliver 2, then enable the flag). Verified it fails against the previous code with `AssertionError: 4.0 != 2.0` and passes now.

The trade-off: a non-detailed pack that moves nothing simply stays at `Delivered = 0`. That is intentional — such a pack shouldn't be invoiced on delivery in the first place, that is what `invoice_policy` is for.

gal-adhoc: Agreed, removed. You were right that it shouldn't be there — and it turned out to be the visible edge of a bigger problem: it only existed to feed a block that wrote `qty_delivered = product_uom_qty` on confirmation. @dcanete-pyming then reproduced the concrete damage on Runboat (delivered quantity restated from 2 to 4 after enabling the flag on an already-delivered order).

I removed that block entirely, so this depend is gone along with it.

sale_stock_product_pack/models/sale_order.py
gal-adhoc: Agreed, removed as well. It was only needed by the `dont_create_move` block that assigned `qty_delivered` from it directly — with that block gone, the `pack_child_line_ids` depends are sufficient, exactly as you said.

Worth noting for the record: this one had been *added* at another reviewer's request earlier in the PR ([here](https://github.com/OCA/product-pack/pull/262#discussion_r3579786171)) to fix a stale-value case in that same block. Removing the block resolves both concerns at once.

sale_stock_product_pack/models/sale_order.py
gal-adhoc: Done — the comment now states these are the same depends `sale_stock` declares for `qty_delivered`, with a reference to the upstream method. The remaining four are exactly the ones you listed.

gal-adhoc:
## Change of approach after review

Thanks @victoralmau @dcanete-pyming @Sadiq-OSI — the reviews converged on the same place, so I reworked the PR rather than patching around it.

**What I removed:** a block in `_compute_qty_delivered` that wrote `qty_delivered = product_uom_qty` on confirmation for packs with `dont_create_move`. Every issue raised in this PR traced back to it:

| Reported | By |
|---|---|
| Delivered quantity restated retroactively (2 → 4) after enabling the flag | @dcanete-pyming |
| `product_id.dont_create_move` shouldn't be in `depends` | @victoralmau |
| `product_uom_qty` shouldn't be in `depends` | @victoralmau |
| Quantity change after confirmation not recomputed | @Sadiq-OSI |
| Draft quotation marked as delivered | @Sadiq-OSI |

Writing a stock-driven field from order state was the wrong mechanism — the two `depends` @victoralmau objected to existed *only* to feed that block, and the two cases @Sadiq-OSI reported were symptoms of it. Removing it resolves all five, and the `depends` list is now exactly the four @victoralmau pointed at.

**What the PR keeps** (the actual fix for the reported issue): `stock.rule.run()` no longer ignores `dont_create_move` for non detailed packs, plus the components-level `depends` so a pack's `qty_delivered` is recomputed when its components are returned.

**Deliberate trade-off:** a non detailed pack that moves nothing now stays at `Delivered = 0`. That is the semantically correct outcome — such a pack shouldn't be invoiced on delivery, which is what `invoice_policy` is for. Note this does contradict `dont_create_move`'s current help text ("will be set as delivered upon sale confirmation"); happy to submit a follow-up updating that text if you agree the field shouldn't be promising to write a stock field.

**Tests:** 29 passing across `product_pack` / `sale_product_pack` / `stock_product_pack` / `sale_stock_product_pack`. Both new regression tests were verified to fail against the previous code (`1.0 != 0.0` and `4.0 != 2.0` respectively).

victoralmau (approved):
In my opinion, the change makes sense.

PR #264, opened 2026-07-24
[18.0][IMP] product_pack: archive pack lines when their product is archived

miquelrosell99:
This PR adds an \active\ field on \product.pack.line\ that is stored, readonly, and computed from both the parent pack product and the component product.

When a product is archived, its pack lines become inactive automatically. Because Odoo applies the standard active test on One2many relations, archived components no longer appear in \product.pack_line_ids\ unless inactive records are explicitly requested.

This fixes the common issue where archived products continue to show up in reports, sale orders, or any other place that lists the contents of a pack.

**Changes:**
- Add computed stored \active\ field on \product.pack.line\.
- Add tests covering component and parent product archival.

**Use case:**
A product that is no longer sold/manufactured is archived, but it is still referenced as a component in existing packs. Without this change it keeps appearing in pack listings.

---
I confirm I have signed the CLA and read the PR guidelines at www.odoo.com/submit-pr

### DIFF
--- a/product_pack/models/product_pack_line.py
+++ b/product_pack/models/product_pack_line.py
@@ -29,6 +29,17 @@ class ProductPackLine(models.Model):
         index=True,
         required=True,
     )
+    active = fields.Boolean(
+        compute="_compute_active",
+        store=True,
+        readonly=True,
+        help="Inactive when the parent or component product is archived.",
+    )
+
+    @api.depends("parent_product_id.active", "product_id.active")
+    def _compute_active(self):
+        for line in self:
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):
LGTM. Makes sense.

Even if I'm wondering how users could be warned about the product archive impact on concerned packs ?

pedrobaeza (changes requested):
I don't get the use case in current module. Where are you seeing the pack lines apart from the product form? And even more, accessing archived packs won't see the pack lines.

